PR #5, opened 2025-10-31
Registry server

ilackarms:
a lot of code is copied directly from the OSS registry and can be cleaned up

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,14 +8,8 @@ on:
 
 jobs:
   build:
-    runs-on: ${{ matrix.os }}
-    strategy:
-      matrix:
-        os: [ubuntu-latest, macos-latest]
-        go-version: [1.22.x]
-        node-version: [18.x]
-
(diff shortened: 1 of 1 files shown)


internal/registry/database/database.go
```diff
@@ -0,0 +1,84 @@
+package database
+
+import (
```
peterj, line 3: there's already a database.go in internal folder -- are we keeping two separate databases?

internal/registry/importer/importer.go
```diff
@@ -0,0 +1,341 @@
...
+}
+
+// SetUpdateIfExists toggles replacing existing name/version entries instead of skipping
+func (s *Service) SetUpdateIfExists(update bool) {
+	s.updateIfExists = update
+}
+
+// SetGitHubToken sets a token used only for GitHub enrichment calls
+func (s *Service) SetGitHubToken(token string) {
+	s.githubToken = strings.TrimSpace(token)
```
peterj, line 61: how will this work practically? are we going to ask the users to provide a github token in order for us to do the enrichment?

PR #18, opened 2025-11-06
Enrich server data with import

ilackarms:


internal/registry/importer/container_scan.go
```diff
@@ -0,0 +1,186 @@
+package importer
```
peterj, line 1: just rememberd -- can you also add the pre-fetched servers.json and have it embedded/load automatically when the daemon starts?

internal/cli/export.go
EItanya, line 1: I'm not sure import/export commands are relevant anymore since the CLI no longer has direct access to the DB. Does this need to happen via the server, or should this just be a separate script?

PR #30, opened 2025-11-08
arctl deploy/remove mcp + UI work to deploy the servers and show the …

peterj:
…live view


internal/registry/database/database.go
```diff
@@ -61,6 +61,18 @@ type AgentFilter struct {
...
+	ServerName   string            `json:"serverName"`
+	Version      string            `json:"version"`
+	DeployedAt   time.Time         `json:"deployedAt"`
+	UpdatedAt    time.Time         `json:"updatedAt"`
+	Status       string            `json:"status"`
+	Config       map[string]string `json:"config"`
+	PreferRemote bool              `json:"preferRemote"`
+	ResourceType string            `json:"resourceType"` // "mcp" or "agent"
+}
+
```
EItanya: This should be in the `models` package. That's where we've been keeping the types.

internal/registry/api/handlers/v0/deployments.go
```diff
@@ -0,0 +1,206 @@
...
+		}
+
+		return &DeploymentResponse{Body: *deployment}, nil
+	})
+
+	// Update deployment configuration
+	huma.Register(api, huma.Operation{
+		OperationID: "update-deployment-config",
+		Method:      http.MethodPut,
+		Path:        basePath + "/deployments/{serverName}/config",
```
EItanya: Why `/config`? I think doing a `PUT` on the same path indicates an update

internal/registry/service/registry_service.go
```diff
@@ -31,6 +32,11 @@ func NewRegistryService(db database.Database, cfg *config.Config) RegistryServic
 	}
 }
 
+// SetReconciler sets the reconciler for server-side container management
+func (s *registryServiceImpl) SetReconciler(reconciler Reconciler) {
+	s.reconciler = reconciler
+}
```
EItanya: Why is this not done at `init`?

internal/registry/service/registry_service.go
```diff
@@ -540,3 +546,113 @@ func (s *registryServiceImpl) createAgentInTransaction(ctx context.Context, tx p
...
+	if config == nil {
+		deployment.Config = make(map[string]string)
+	}
+
+	err = s.db.CreateDeployment(ctx, nil, deployment)
+	if err != nil {
+		return nil, err
+	}
+
+	// Trigger reconciliation if available (runs in background)
```
EItanya: Why in background? I think we need to know if this fails. Or at the very least we need to indicate in the database object that this is just "starting" and that it can poll to see when it's done. For now I think sync is probably best.

internal/registry/registry_app.go
```diff
@@ -101,25 +114,47 @@ func App(_ context.Context) error {
...
 	}()
 
+	if cfg.ReconcileOnStartup && reconciler != nil {
+		go func() {
+			<-serverStarted
+
+			maxRetries := 10
+			retryDelay := 500 * time.Millisecond
+
+			log.Println("Waiting for database to be ready before reconciling...")
```
EItanya: Can you just ping the DB to make sure, honestly the app shouldn't start up at all unless it's available. For example, on line 51 right after you get the connection, just send a ping to make sure it's alive. Or alternatively, can we make sure that this container doesn't start until the other one is available, does docker-compose support something like that.

peterj: fixed, no need to ping as it's already done in the newpostgresql

PR #34, opened 2025-11-08
mcp init / build / add-tool

ilackarms:


peterj (approved):

internal/cli/mcp/frameworks/golang/templates/go.mod.tmpl
```diff
@@ -0,0 +1,7 @@
+module {{.GoModuleName}}
+
+go 1.23.0
```
EItanya, line 3: Definitely not a blocker but we should upgrade this

EItanya (commented):
My one question is what's the difference/relationship between an `mcp.yaml` and an `mcp.json` and how are we going to explain that to users?

internal/cli/mcp/build/build.go
```diff
@@ -0,0 +1,176 @@
+package build
+
```
peterj, line 2: kmcp has the run command too, right? can we add that one here as well -- ideally, we'd invoke it from the `run.go` -- we have code there that can run an mcp server from the registry, but we can check whether the folder has mcp.yaml/mcp.json and instead invoke the kmcp's run code for now.

ilackarms:
> My one question is what's the difference/relationship between an `mcp.yaml` and an `mcp.json` and how are we going to explain that to users?

not sure what you mean? what's mcp.json?

internal/cli/mcp/build/build.go
ilackarms: sure, i can do a follow up or add to this pr

EItanya (approved):

PR #35, opened 2025-11-09
admin endpoints, mark items as "published=false" when imported

peterj:
add publish commads + fix up the UI to use the end endpoints.


internal/registry/api/handlers/v0/servers.go
```diff
@@ -53,19 +54,32 @@ type ServerReadmeResponse struct {
 }
 
 // RegisterServersEndpoints registers all server-related endpoints with a custom path prefix
-func RegisterServersEndpoints(api huma.API, pathPrefix string, registry service.RegistryService) {
+// isAdmin: if true, shows all resources; if false, only shows published resources
+func RegisterServersEndpoints(api huma.API, pathPrefix string, registry service.RegistryService, isAdmin bool) {
+	// Determine the tag based on whether this is admin or public
+	tag := "servers"
+	if isAdmin {
+		tag = "admin"
```
ilackarms: shouldnt admin be an additional tag? why do we replace the original tag which seems relevant

peterj: good point, will just append admin for admin stuff

PR #46, opened 2025-11-11
add --platform to skill publish

peterj:


EItanya:
Oh this is annoying, and maybe a reason not to use docker. The architecture here is really irrelevant because it's just a bunch of files. Most likely we should try to always build multi-arch, but that may be a little harder

PR #59, opened 2025-11-19
(CLI) Only require agentregistry to be running for commands that require it

### LINKED ISSUE #31: agent init shouldn't depend on the registry running
```
Error: API client not initialized: failed to reach API at http://localhost:12121/v0: Get "http://localhost:12121/v0/ping": read tcp [::1]:51427->[::1]:12121: read: connection reset by peer
Usage:
  arctl agent init [framework] [language] [agent-name] [fl
```

inFocus7:
**work in progress**

note: initially removed the `apiClient` global definitions because the previous setup was _always_ setting at-start (which is what caused failures when trying to do simple commands like `version`). we're moving away from that because not all require it; the api client needs to be set on the **verb** level (`root noun verb`), so we can be explicit on which commands it's needed for. while this works, this means passing an `apiClient` argument around to the helpers, which is okay, but i'm not the biggest fan of how much it's being passed around. i'm going to look into an alternative to this.

# Changes

- Remove `root` preRunCheck to avoid registry check run on a per-command basis.
- Explicitly add checks for the registry & set apiClient where necessary.

## TODO

- [ ] look into if there is a long-running interactive mode where we would want the apiClient configured globally/noun-level instead of _always_ (re)fetched on the per-verb level.
- [ ] manual validation.
- [ ] automated tests to validate cli behavior(?)

# Context

Currently when running any command, we run checks for the registry to be running. We **do not** require the registry to run to handle commands such as `init` -- which simply handles templating. These changes make sure we only do the checks when required.

# Note(s)

We had to explicitly move the check into (sub)commands that require it because the structure is `noun verb`, where we only require the registry running based on `verbs`, which is the most granular subcommand.

**If** the CLI was set up in a `verb noun` pattern (e.g. `kagent init <agent,mcp>`), then it would be _less_ explicit where we would have only required setting the check on the root verb.

Resolves #31

### DIFF
--- a/internal/cli/mcp/deploy.go
+++ b/internal/cli/mcp/deploy.go
@@ -83,7 +83,7 @@ func runDeploy(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("server not found: %s", serverName)
 	}
 
-	isPublished, err := isServerPublished(serverName, deployVersion)
+	isPublished, err := isServerPublished(apiClient, serverName, deployVersion)
 	if err != nil {
 		return fmt.Errorf("failed to check if server is published: %w", err)
 	}
--- a/internal/cli/mcp/helpers.go
+++ b/internal/cli/mcp/helpers.go
@@ -7,10 +7,11 @@ import (
 	"os"
 	"strings"
 
+	"github.com/agentregistry-dev/agentregistry/internal/client"
 	apiv0 "github.com/modelcontextprotocol/registry/pkg/api/v0"
 )
 
-func isServerDeployed(serverName string, version string) (bool, error) {
+func isServerDeployed(apiClient *client.Client, serverName string, version string) (bool, error) {
 	if apiClient == nil {
 		return false, errors.New("API client not initialized")
 	}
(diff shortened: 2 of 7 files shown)


internal/cli/agent/agent.go
```diff
@@ -1,16 +1,10 @@
 package agent
 
 import (
-	"github.com/agentregistry-dev/agentregistry/internal/client"
```
peterj, line 4: just a note that @ilackarms is working on bringing in the code from kagent (just want to avoid huge merge conflicts).

inFocus7: Ah alright! I'll hold off on these changes until after the kagent code is brought in 🫡

peterj:
at least changes in the /agent folder :)

inFocus7:
> at least changes in the /agent folder :)

Yeah I was thinking about waiting vs. undoing portions where there could be nasty conflicts, but leaned towards waiting and resolving the conflicts myself after Scott's work.

If I make these changes excluding /agent to lessen conflicts for Scotts work, there are two options:
A. Remove the `root`-level checks --> agent commands will no longer have checks, although the non-agent commands will work as expected.
B. Keep the `root`-level checks --> agent commands will keep the checks, **but** we'd be double checking in the other sub-commands. This shouldn't be breaking, but not ideal.

---

Do you know if there's an ETA for that work? I could also merge in one of the two options above (likely **B**), then do a follow-up to clean it up + handle it in /agent. The above likely isn't an issue as long as there isn't a release in the in-between.

PR #68, opened 2025-11-25
implement publish/unpublish/delete commands

peterj:


internal/cli/agent/delete.go
```diff
@@ -0,0 +1,109 @@
...
+
+	return false, nil
+}
+
+func isAgentDeployed(agentName, version string) (bool, error) {
+	if apiClient == nil {
+		return false, fmt.Errorf("API client not initialized")
+	}
+
+	deployment, err := apiClient.GetDeployedServerByNameAndVersion(agentName, version)
```
EItanya, line 100: Isn't `DeployedServer` a tool?

internal/cli/agent/delete.go
```diff
@@ -0,0 +1,109 @@
...
+func runDelete(cmd *cobra.Command, args []string) error {
+	agentName := args[0]
+
+	if apiClient == nil {
+		return fmt.Errorf("API client not initialized")
+	}
+
+	if deleteVersion == "" {
+		return fmt.Errorf("version is required (use --version flag)")
+	}
```
EItanya: You mark the flag required, this can never happen

EItanya (approved):

internal/cli/agent/delete.go
peterj: I don't understand the question - deployment is a deployment of an MCP server

PR #69, opened 2025-11-26
Migrate `arctl agent add-mcp` from kagent + implement `--from-registry` MCP type

inFocus7:
image

In AgentRegistry persona terminology: `--from-registry` is for a consumer of registry published artifacts, since the manifest would store a pointer that to the registry resource, which will be pulled as-needed (runtime). For runtime registry consumption (i.e. calling mcp deployed elsewhere), _I believe_ it should be already possible to set the `remote` mcp type referencing the agentregistry.

# Changes

- [**done**] migrate add-mcp command + wizard ui from kagent to agentregistry
- [**done**] added new registry type for agent manifest
- [**done**] registry type resolution when running `agent run`
    - [**done**] agent local directory:
        - creates files as normal
        - creates agw + docker setup for registry-resolved
        - stores deploy info under `registry/` so its easy to clean & re-create per-run, avoiding conflicts.
        - creates config file w/ resolved agents to inject to agent (instead of baking into `mcp_tools.py`, to be consistent with our other running methods)
    - [**done**] agent registry:
        - requires built images for dockercompose, as normal
        - registry-resolved servers resolve at runtime so we build those images + add to docker compose
        - creates config file w/ resolved agents to inject to agent (this relies on pre-built agent, which won't have resolved agents in their mcp_tools.py)
- [**done**] set up registry resolution when going through `agent build,publish,deploy` pipeline
    - a caveat is the `registryUrl`, running an agent locally should be `127.0.0.1:12121`, **but** when deploying within the registry, I needed to use `host.docker.internal` since `127.0.0.1:12121` wasn't resolved. (maybe `127.0.0.1:8080` would have worked?)

## Note

In order to support resolving mcp servers at run/deploy-time, we'll need to mount their configuration to pre-built agent images.

When deploying, for instance, a pre-built agent already exists without resolved servers, meaning we can't bake-in the mcp server configs in the `mcp_tools.py` directly, therefore we inject that data through the mount.

This led to some extra things needed:
1. Mounting config file(s) - just the generated `mcp-servers.json` file, for now at least
2. Passing around the agent `version`.
    - Because multiple versions of an agent could be deployed, we need to know agent/version to store & mount mcp configs.
3. Edge case handling.
    - If an agent manifest goes from having a registry server defined to _not_ defining, we need to ensure the previous mcp-servers.json generated isn't re-used. In this case, during deployments, if no registries were resolved, we delete any pre-existing config. While this isn't currently possible, and may never be, it's still safer to handle.

Note: The below **todo**, at least the `arctl agent add-mcp --init` will be a follow-up. This PR is already sizable, and I want to avoid adding more to it if not necessary.
- [**todo**] allow to run from *.yaml/*.json file created from running `arctl mcp init` within an agent project dir.
    - note: should have an `agent add-mcp --init` flag so that it _knows_ to add it to compose correctly, init should follow the same path as `mcp init`.
- [**todo**] look into if there are e2e & integration tests to have a grander testing suite for these changes. 

# Validation Steps

## Setup Local Environment

```sh
docker volume rm agentregistry_postgres_data

make docker-registry docker docker-compose-up
```

## Create Agent

```sh
# init
go run cmd/cli/main.go agent init adk python test --model-provider OpenAI --model-name gpt-5-mini
```

## Add MCP Servers

```sh
# From the UI, publish an MCP server.
# note: I used the pokemon api server
open http://localhost:12121
```

```sh
# command mcp
go run cmd/cli/main.go agent add-mcp everything --project-dir test/ --command npx --arg @modelcontextprotocol/server-everything

# mcp from registry (new)
# todo: this registry url is tricky - running locally, you should point to `localhost/127.0.0.1`, but when publishing, should use `host.docker.internal` so it resolves itself correctly within the registry service
go run cmd/cli/main.go agent add-mcp poke --project-dir test/ --registry-url http://localhost:12121 --registry-server-name io.github.Asthanaji05/pokeapi-mcp-server --registry-server-version 1.9.0

# TODO: init an mcp within the agent and reference by its *.yaml file (may need to add support)
```

## Validate Agent

```sh
# below should list everything, including a `registry` 
cat test/agent.yaml | yq .mcpServers
```

## Build Agent

```sh
go run cmd/cli/main.go agent build test/ --verbose
```

```sh
docker images | grep test
```

## Run Agent

_note: i hit issues previously due to agent name in agent.py being invalid (`test_agent`) (happened on `run` where it was a logged problem in service, and on-deploy which responded with a 400 due to naming)

we should probably do validation on `agent init` so it follows proper naming conventions and users don't hit this AFTER configuring everything.

```sh
go run cmd/cli/main.go agent run ./test
```

## Push Agent to Registry

```sh
go run cmd/cli/main.go agent publish test/
```

_Note: This only pushes, you'll need to publish through the admin UI_

## Run Registry Agent

```sh
go run cmd/cli/main.go agent run test
```

## Deploy

```sh
go run cmd/cli/main.go agent deploy test
```

### Deployment Validation

```sh
cd /tmp/arctl-runtime
```

```sh
# check that the setup for remote registry is binded
cat /tmp/arctl-runtime/mcp-servers-test.json
```

```sh
# check that non-resolved (remote, command) are baked-in
docker compose exec test cat /app/test/mcp_tools.py
```

```sh
# todo; why is the command-based server missing?
cat /tmp/arctl-runtime/agent-gateway.yaml
```

# Debugging

Once above is running, while testing, you can run the below to rebuild & run the server(s)

```sh
make docker-server docker-compose-up
```

```sh
make docker-agentgateway docker-compose-up
```

# Demos

wip

**wizard**

https://github.com/user-attachments/assets/199576ae-4f48-4828-9cd5-e6b92cea53d9

**local agent directory run**

https://github.com/user-attachments/assets/8bf5e671-8e16-4a9b-ac22-60ee8a5b5bc1

**registry agent run** (+ deploy)

https://github.com/user-attachments/assets/e8a7a2d2-8b27-43b7-ab4d-8b2de31397f1

_note: deploy works similarly, where it injects resolved registry-based mcps (although it's done within the registry) so I didn't record a separate video._

_note2: these videos were recording before a few cleanups + a change in local directory run (which now injects additional mcps through config, similar to deploys and registry-agent-runs)_

### DIFF
--- a/internal/cli/agent/frameworks/common/mcp_config.go
+++ b/internal/cli/agent/frameworks/common/mcp_config.go
@@ -1,0 +1,108 @@
+package common
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/agentregistry-dev/agentregistry/internal/utils"
+)
+
+// MCPConfigTarget describes where to place the resolved MCP config for an agent.
+type MCPConfigTarget struct {
+	BaseDir   string
+	AgentName string
+	Version   string // optional; when set, path includes sanitized version
+}
+
+// ComputeMCPConfigPath returns the config directory and file path for an agent.
+// If BaseDir or AgentName is empty, both returned paths are empty.
+func ComputeMCPConfigPath(target *MCPConfigTarget) (configDir string, configPath string) {
+	if target.BaseDir == "" || target.AgentName == "" {
+		return "", ""
+	}
+
+	configDir = filepath.Join(target.BaseDir, target.AgentName)
+	if target.Version != "" {
+		configDir = filepath.Join(configDir, utils.SanitizeVersion(target.Version))
+	}
+	configPath = filepath.Join(configDir, "mcp-servers.json")
+	return configDir, configPath
+}
+
+// RefreshMCPConfig cleans any existing MCP config for the target and optionally writes a new one.
+// If servers is empty or nil, it performs cleanup only.
+func RefreshMCPConfig(target *MCPConfigTarget, servers []PythonMCPServer, verbose bool) error {
+	if target == nil {
+		return fmt.Errorf("target is required")
+	}
+
+	configDir, configPath := ComputeMCPConfigPath(target)
+	if configDir == "" {
+		return nil
+	}
+
+	if err := os.Remove(configPath); err != nil && !os.IsNotExist(err) {
+		return fmt.Errorf("failed to remove MCP server config file: %w", err)
+	}
+
+	// Attempt to remove the directory if it is empty; ignore errors.
+	_ = os.Remove(configDir)
+
+	if len(servers) == 0 {
+		return nil
+	}
+
+	if err := os.MkdirAll(configDir, 0755); err != nil {
+		return fmt.Errorf("failed to create agent config directory: %w", err)
+	}
+
+	configData, err := json.MarshalIndent(servers, "", "  ")
+	if err != nil {
+		return fmt.Errorf("failed to marshal MCP server config: %w", err)
+	}
+
+	if err := os.WriteFile(configPath, configData, 0644); err != nil {
+		return fmt.Errorf("failed to write MCP server config file: %w", err)
+	}
+
+	if verbose {
+		fmt.Printf("Wrote MCP server config for agent %s version %s to %s\n", target.AgentName, target.Version, configPath)
+	}
+
+	return nil
+}
+
+// PythonServersFromManifest converts resolved MCP servers in a manifest into Python MCP server structs.
+// Registry-type servers are skipped because they are not directly runnable.
+func PythonServersFromManifest(manifest *AgentManifest) []PythonMCPServer {
+	if manifest == nil || len(manifest.McpServers) == 0 {
+		return nil
+	}
+
+	var mcpServers []PythonMCPServer
+	for _, srv := range manifest.McpServers {
+		if srv.Type == "registry" {
+			continue
+		}
+
+		pythonServer := PythonMCPServer{
+			Name: srv.Name,
+			Type: srv.Type,
+		}
+
+		if srv.Type == "remote" {
+			pythonServer.URL = srv.URL
+			if len(srv.Headers) > 0 {
+				pythonServer.Headers = srv.Headers
+			}
+		}
+		// For command type, the Python code constructs URL as f"http://{server_name}:3000/mcp"
+		// so we do not set URL here.
+
+		mcpServers = append(mcpServers, pythonServer)
+	}
+
+	return mcpServers
+}
--- a/internal/cli/agent/project/project.go
+++ b/internal/cli/agent/project/project.go
@@ -198,15 +198,8 @@ func EnsureMcpServerDirectories(projectDir string, manifest *common.AgentManifes
 
 	// Clean up registry/ folder to ensure fresh state for registry-resolved servers.
 	// This prevents stale configs from previous runs with different resolved registries.
-	registryDir := filepath.Join(projectDir, "registry")
-	if _, err := os.Stat(registryDir); err == nil {
-		if err := os.RemoveAll(registryDir); err != nil {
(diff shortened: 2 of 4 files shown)


internal/cli/agent/utils/registry_translator.go
```diff
@@ -0,0 +1,76 @@
...
+	"fmt"
+
+	"github.com/agentregistry-dev/agentregistry/internal/cli/agent/frameworks/common"
+	"github.com/agentregistry-dev/agentregistry/internal/registry/types"
+	"github.com/agentregistry-dev/agentregistry/internal/runtime/translation/registry/utils"
+)
+
+// TranslateRegistryServer converts a registry ServerSpec into a common.McpServerType
+// that can be used by the docker-compose generator.
+// TODO: What are cases where there are both remotes and packages? Or multiple of any?
```
peterj: I think we have a boolean somewher called "preferRemote" where the user can specify whether they prefer the remote vs. the package option (in case both exist).

internal/cli/agent/utils/registry_translator.go
```diff
@@ -0,0 +1,76 @@
...
+		return &common.McpServerType{
+			Type:    "remote",
+			Name:    name,
+			URL:     remote.URL,
+			Headers: headers,
+		}, nil
+	}
+
+	// if there are packages (command-based servers), use the first one
+	if len(serverSpec.Packages) > 0 {
```
peterj: we can figure this out later -- in case there are multiple package, give user the option to pick the one they want.

internal/cli/agent/utils/registry_translator.go
inFocus7: Yeah I saw that for mcps deployments we use that, although it doesn't seem like we use it for agents ([ref](https://github.com/agentregistry-dev/agentregistry/blob/133e9345a6bb108c2b5654d8a318a21b87f0fa01/internal/registry/service/registry_service.go#L758-L810))

Should the UX be:
A. Agent-level sets `preferRemote` (unsure where it's done, but judging by the code, it's possible), this then gets applied to _all_ its servers we resolve (registry-type servers)
B. Define a preferRemote flag on the agent manifest for registry mcp servers, so we can `add-mcp registry --prefer-remote`. this way when we resolve them from the registry, we use the flag.

I believe you brought up possibly doing (B) before, which I think would be the best to implement. I have uncommitted changes locally handling (B) if that's what we decide on 👍🏼

peterj: I think it should be per MCP server  -- so anytime you call add-mcp and the server has a remote and a non-remote package, we should warn the user/tell them to explicitly pick which one they want

inFocus7: I added a `--prefer-remote` flag (technically `--registry-server-prefer-remote`**) for now. 

My concern with having us resolve it during the `add-mcp` phase is to show the packages/command:
1. (minor) we'll need to store this choice in the manifest (remote/package, index [if multiple]) so it can be used during resolution on runs.
2. (major) could there be an edge case where a user can chooses `remote[0]` (or just `remote` for simplicity) during `add-mcp`. Then they publish. Then, a few days later they deploy/run. On that deploy/run, if the re-resolved mcp differs we'll need to figure out a way to cleanly handle this. (how to alert the user + how a user can update their preference based on latest resolution [would need ability to update the agent manifest stored in the db] 

I think this choice would fit best when resolving (`run`/`deploy`), although would need to think about the UX a bit more.
- CLI `run`: interactive mode would be easy since it would be a sequential wizard step, flag-based would need a bit more thinking.
- Dashboard `deploy`: Clicking deploy would need a second step where it shows packages & remotes to choose from

** it's a really long name, i could do `--prefer-remote` which is cleaner; the only issue is that it's solely for registry-type servers, so the current name is more explicit.

---

We talked about this during meeting, for now doing as we do now: use the first package/remote, allow for `--prefer-remote`.

Relevant Issue: https://github.com/agentregistry-dev/agentregistry/issues/74

PR #71, opened 2025-12-03
Add `agent chat` for chatting with deployed agents

### LINKED ISSUE #72: Implement `agent chat`
Implement a way for users using the CLI to chat with a **deployed** agent. This is already support for locally-running agents, and it would be nice to have an easy method to chat with agents deployed in agentregistry.

Requirements:
- `arctl agent chat [agent-name] --version [optional-version] --url [a2a url, default to gateway]` -> Start chat, similar to chatting with locally-running agents.
- `arctl agent chat` -> Start interactive selection, displaying agents + versions based on selection & url, after selection start a chat similar to chatting with locally-running agents.

inFocus7:
**paused: waiting for #69 to merge as it has added util setups, that I'd like to leverage here (if possible); don't want to hit merge conflicts. it also has a cool tui/bubbletea wizard brought over from kagent, which i'm going to copy the architecture for the interactive agent+version selection wizard**

note to self for tui installation:
- 4 step process
    - 1: gateway url (for comms). optional, default to the http://localhost:21212. string input.
    - 2. select agent from registry DEPLOYED agent list. list choice.
    - 3. select version of selected agent. list choice.
    - 4. chat!

# Context

I want to chat with a deployed agent easily.

I noticed there was a difference between a registry-published agent running locally + a deployed agent runs (e.g. deployed agent sets up agent gateway config for stdio servers, local sets up services). Being able to easily test out a deployed agent locally is a good option to have.

We already have tui + a2a setup for CLI chatting when running an agent locally, so we can easily leverage that for this.

# Changes

- [**done**] Add `chat` subcommand for `arctl agent` allowing you to chat with a deployed agent.
- [**done**] Add wizard for selecting deployed agent + version to chat with.

Currently this only allows URL configuration for the `gateway-url` (a2a agent gateway endpoint), and uses the created apiClient set up with default registry. I can add another `registry-url` option in order to allow chatting with agents from other registries. 

# Verification

**demo**

https://github.com/user-attachments/assets/b189534c-7502-480a-8771-f7a04c9a7579

---

Resolves #72

### DIFF
--- a/internal/cli/agent/chat.go
+++ b/internal/cli/agent/chat.go
@@ -3,66 +3,79 @@ package agent
 import (
 	"context"
 	"fmt"
-	"os"
 	"time"
 
 	"github.com/agentregistry-dev/agentregistry/internal/cli/agent/tui"
 	"github.com/agentregistry-dev/agentregistry/internal/client"
+	tea "github.com/charmbracelet/bubbletea"
 	"github.com/spf13/cobra"
 	a2aclient "trpc.group/trpc-go/trpc-a2a-go/client"
 	"trpc.group/trpc-go/trpc-a2a-go/protocol"
 )
 
+// NOTE: This command uses the global apiClient instance which uses the default registry (:12121)
+// In order to support custom registries, we'll need to pass it in and create a new client for it.
+// Gateway URL: Agent Gateway, for communicating with deployed agents.
+// Registry URL: Registry, for fetching agent metadata.
+
 var ChatCmd = &cobra.Command{
-	Use:   "chat <deployed-agent-name>",
+	Use:   "chat [deployed-agent-name]",
 	Short: "Chat with a deployed agent",
-	Long: `Chat with a deployed agent through the agent gateway. If --version is not provided,
-the command will attempt to find the deployed version automatically. If multiple versions
-are deployed, you must specify --version.`,
-	Args: cobra.ExactArgs(1),
+	Long: `Chat with a deployed agent through the agent gateway. If no agent name is provided,
+a wizard will guide you through selecting an agent and version. If --version is not provided
+when specifying an agent name, the command will attempt to find the deployed version automatically.
+If multiple versions are deployed, you must specify --version.`,
+	Args: cobra.MaximumNArgs(1),
 	RunE: runChat,
-	Example: `arctl agent chat my-agent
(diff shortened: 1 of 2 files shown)


internal/cli/agent/chat.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/agentregistry-dev/agentregistry/internal/client"
+	"github.com/spf13/cobra"
+	a2aclient "trpc.group/trpc-go/trpc-a2a-go/client"
+	"trpc.group/trpc-go/trpc-a2a-go/protocol"
+)
+
+var ChatCmd = &cobra.Command{
+	Use:   "chat <deployed-agent-name>",
+	Short: "Chat with a deployed agent",
+	Long: `Chat with a deployed agent through the agent gateway. If --version is not provided,
```
peterj: since `arctl agent run` launches automatically into the chat experience, I am wondering if we keep that same command and handle chatting with a deployed agent through that. 

Comparing how we do this for MCP, we have an option to launch an inspector when doing `arctl mcp run` -- and in that case run handles both local MCP servers and published servers from the registry (it doesn't handle the deployed servers...)

1. If agent is local (code on disk): arctl agent run will build& run the agent and launch into the chat experience
2. If agent is in published in the registry: arctl agent run should run the published agent locally and then launch into the chat 
3. If agent is deployed in the registry: arctl agent run would just launch into the chat experience with the deployed agent. 


What do you think? I might have to sleep on it :D

inFocus7: Yeah, I think that makes sense, we already handle (1) and (2), so this should be easy enough.
I think the only thing is that we'll need aa way to know if an agent has been deployed or just exists in the registry. I haven't checked if we have this API/functionality yet, but just noting (more for myself).

peterj: yeah, I think there is an endpoint to get deployed items by name/version

inFocus7: Actually, i slept on it over a weekend and now I'm pro-`chat` 😅

I think `chat` would be best since this would solely be for chatting, and wouldn't necessarily run anything. The `run` already handles two cases - running agent from directory & running agent from registry, so adding a third purpose ("running" a chat w/ deployed agent) could make it a bit more complex (confusing).

PR #86, opened 2026-01-05
Allow Auth Extensibility

inFocus7:
Exposing middleware hooks + extension points needed for libraries extending agentregistry.

As usual with adding extensibility (like in #83) there are some refactors that are done to move internal objects for external usage.

**TODO** (create issues for):
- Better handling of authz failures in handlers.
    - Currently, for security we return a 404 on resources not found either due to non-existence or authz failure. This is to avoid existence leakage.
    - This only makes sense when registries are public and want added security, but this may not _always_ be the case and could cause confusion, so should rethink this a bit.

## Changes
1. Update list of actions + artifacts to match access control resources
2. Expose an AuthN + AuthZ app extensions for external libraries to define.
3. Create a public (unsafe) authZ in the meantime for all* checks to pass
    - "all", expect **edit**, to match existing behavior.
4. Expose DB extension so external libraries can override specific DB queries to fit needs
    - Additionally: Moved internal DB to external pkg.
5. Call appropriate authz checks in DB handlers; List functionality should be handled through DB filtering to avoid N authz.Checks().
6. Added CLI authN extension to allow configurable authn with CLI commands.
6. Fixed deployment methods to _specifically_ check queried resources.
    - From what I understood, we had half-baked functionality before, where we didn't do some important things like checking the exact resource (based on type > name > version) by omitting the type or version.
8. Update the publish page to show 'already deployed' for deployed agents to match servers.
    - We currently allow clicking deploy and then returns toaster stating it's deployed, so this is a slightly better UX.
9. Updated the new mcp server auth to use new AuthN/AuthZ configuration (instead of solely the JWTManager)

For privacy, to prevent existence leakage, any forbidden response in auth returns a 404.

## Notes

- There are three unsafe-ish CLI commands: `import`,`export`,`embeddings`. These communicate with the DB directly instead of going through the API which others do. This means they don't go through the authN middleware that stores the user session in-context. In the meantime, I disabled the authz checks for these (`{authz: nil}`), but we'll want to update these commands to follow the same path as the other CLI commands for porper authz.

**Important**

The database extension assumes that extension db migrations are additions - extension tables-only - in which the ordering does not matter, since they wouldn't need to interleave with OSS' tables.

All OSS migrations are ran first, then extension migrations. If an extension **does** require interleaving (e.g. ext uses oss tables/rows), then there could be an issue. 

For example, in the following order of events:
1. 001_OSS.sql: Create table
2. 00N_EXT.sql: Uses table row as FK.
3. 002_OSS.sql: Remove table row
5. 00N+1_EXT: Remove usage as FK

New setups would do: [001_OSS, 002_OSS, 00N_EXT, 00N+1_EXT], where 00N_EXT would reference an inexistent table and fail.

It may be worth syncing up with others to spitball ideas (e.g., timestamp-based migration files - which can have its own set of issues, migration dependency setup).

image

### DIFF
--- a/internal/cli/agent/utils/registry_resolver.go
+++ b/internal/cli/agent/utils/registry_resolver.go
@@ -11,6 +11,16 @@ import (
 	"github.com/modelcontextprotocol/registry/pkg/model"
 )
 
+var defaultRegistryURL = "http://127.0.0.1:12121"
+
+// SetDefaultRegistryURL overrides the fallback registry URL used when manifests omit registry_url.
+func SetDefaultRegistryURL(url string) {
+	if strings.TrimSpace(url) == "" {
+		return
+	}
+	defaultRegistryURL = url
+}
+
 // ParseAgentManifestServers resolves registry-type MCP servers in an agent manifest, keeping non-registry servers as-is.
 func ParseAgentManifestServers(manifest *models.AgentManifest, verbose bool) ([]models.McpServerType, error) {
 	servers := []models.McpServerType{}
@@ -58,7 +68,7 @@ func ParseAgentManifestServers(manifest *models.AgentManifest, verbose bool) ([]
 func resolveRegistryServer(mcpServer models.McpServerType, verbose bool) (*models.McpServerType, error) {
 	registryURL := mcpServer.RegistryURL
 	if registryURL == "" {
-		registryURL = "http://127.0.0.1:12121"
+		registryURL = defaultRegistryURL
 		if verbose {
 			fmt.Printf("[registry-resolver]   Using default registry URL: %s\n", registryURL)
 		}
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -27,34 +27,12 @@ type Client struct {
 const (
 	defaultRegistryName = "local"
 	defaultBaseURL      = "http://localhost:12121/v0"
+	DefaultBaseURL      = defaultBaseURL
 )
 
 // NewClientFromEnv constructs a client using environment variables
 func NewClientFromEnv() (*Client, error) {
-	base := os.Getenv("ARCTL_API_BASE_URL")
-	if strings.TrimSpace(base) == "" {
-		base = defaultBaseURL
-	}
-	token := os.Getenv("ARCTL_API_TOKEN")
-	c := &Client{
-		BaseURL: base,
-		token:   token,
-		httpClient: &http.Client{
-			Timeout: 30 * time.Second,
-		},
-	}
-	// Verify connectivity
-	// retry backoff with exponential backoff
-	for i := range 5 {
-		if err := c.Ping(); err != nil {
-			if i == 2 {
-				return nil, fmt.Errorf("failed to reach API after 3 attempts: %w", err)
-			}
-			time.Sleep(time.Duration(i+1) * time.Second)
-		}
-	}
-	// Seed placeholder registry entry
-	return c, nil
+	return NewClientWithConfig(os.Getenv("ARCTL_API_BASE_URL"), os.Getenv("ARCTL_API_TOKEN"))
 }
 
 // NewClient constructs a client with explicit baseURL and token
@@ -69,6 +47,35 @@ func NewClient(baseURL, token string) *Client {
 			Timeout: 30 * time.Second,
 		},
 	}
+}
+
+// NewClientWithConfig constructs a client from explicit inputs (flag/env), applies defaults, and verifies connectivity.
+func NewClientWithConfig(baseURL, token string) (*Client, error) {
+	base := strings.TrimSpace(baseURL)
+	if base == "" {
+		base = defaultBaseURL
+	}
+
+	c := NewClient(base, token)
+	if err := pingWithRetry(c); err != nil {
+		return nil, fmt.Errorf("failed to reach API at %s: %w", c.BaseURL, err)
+	}
+
+	return c, nil
+}
+
+func pingWithRetry(c *Client) error {
+	var lastErr error
+	const attempts = 3
+	for i := range attempts {
+		if err := c.Ping(); err != nil {
+			lastErr = err
+			time.Sleep(time.Duration(i+1) * time.Second)
+			continue
+		}
+		return nil
+	}
+	return fmt.Errorf("failed to reach API after %d attempts: %w", attempts, lastErr)
 }
 
 // Close is a no-op in API mode
--- a/pkg/cli/root.go
+++ b/pkg/cli/root.go
@@ -2,10 +2,13 @@ package cli
 
 import (
 	"fmt"
+	"net/url"
 	"os"
+	"strings"
 
 	"github.com/agentregistry-dev/agentregistry/internal/cli"
 	"github.com/agentregistry-dev/agentregistry/internal/cli/agent"
+	agentutils "github.com/agentregistry-dev/agentregistry/internal/cli/agent/utils"
 	"github.com/agentregistry-dev/agentregistry/internal/cli/configure"
 	"github.com/agentregistry-dev/agentregistry/internal/cli/mcp"
 	"github.com/agentregistry-dev/agentregistry/internal/cli/skill"
@@ -28,6 +31,10 @@ type CLIOptions struct {
 }
 
 var cliOptions CLIOptions
+var registryURL string
+var registryToken string
+
+const defaultRegistryPort = "12121"
 
 // Configure applies options to the root command
 func Configure(opts CLIOptions) {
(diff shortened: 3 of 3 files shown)


internal/cli/import.go
```diff
@@ -50,7 +51,11 @@ var ImportCmd = &cobra.Command{
 		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 		defer cancel()
 
-		db, err := database.NewPostgreSQL(ctx, cfg.DatabaseURL)
+		// TODO: instead of communicating with db directly, we should communicate through the registry service
```
inFocus7, line 54: not a todo for me, since out of scope; but some commands (import, export, _maybe_ embedding) don't communicate through http handlers. this means the handler middleware that does token extraction + context storage does not take effect; therefore there the db-level authz can't get the token to authz

the other commands should be ok

EItanya: Absolutely, this needs to move

EItanya: It's the same with embedding now

internal/mcp/registryserver/server.go
```diff
@@ -42,37 +34,12 @@ func NewServer(cfg *config.Config, registry service.RegistryService) *mcp.Server
...
 	addServerTools(server, registry)
 	addSkillTools(server, registry)
-	addDeploymentTools(server, registry, jwtManager)
+	addDeploymentTools(server, registry)
 	addMetaTools(server)
 
 	return server
 }
 
-func requireAuthToken(ctx context.Context, jwtManager *auth.JWTManager, resource string) error {
```
EItanya, line 51: I remember chatting about removing this, but I wanted to make sure that the claims still wind up in the `ctx`. That's the only way the authz library will get them

internal/registry/api/handlers/v0/auth/none.go
```diff
@@ -60,6 +60,18 @@ func RegisterNoneEndpoint(api huma.API, pathPrefix string, cfg *config.Config) {
...
+		{
+			Action:          auth.PermissionActionRead,
+			ResourcePattern: "io.modelcontextprotocol.anonymous/*",
+		},
+		{
+			Action:          auth.PermissionActionPull,
+			ResourcePattern: "io.modelcontextprotocol.anonymous/*",
+		},
+		{
+			Action:          auth.PermissionActionPush,
```
EItanya, line 68: What's the difference between `push` and `publish`

internal/registry/api/handlers/v0/auth/none.go
```diff
@@ -60,6 +60,18 @@ func RegisterNoneEndpoint(api huma.API, pathPrefix string, cfg *config.Config) {
 func (h *NoneHandler) GetAnonymousToken(ctx context.Context) (*auth.TokenResponse, error) {
 	// Build permissions for anonymous namespace only
 	permissions := []auth.Permission{
+		{
+			Action:          auth.PermissionActionRead,
+			ResourcePattern: "io.modelcontextprotocol.anonymous/*",
+		},
+		{
+			Action:          auth.PermissionActionPull,
+			ResourcePattern: "io.modelcontextprotocol.anonymous/*",
```
EItanya: What's the difference between `pull` and `read`

internal/registry/api/handlers/v0/auth/none.go
```diff
@@ -68,6 +80,18 @@ func (h *NoneHandler) GetAnonymousToken(ctx context.Context) (*auth.TokenRespons
...
+		},
+		{
+			Action:          auth.PermissionActionDeploy,
+			ResourcePattern: "io.modelcontextprotocol.anonymous/*",
+		},
+		{
+			Action:          auth.PermissionActionRun,
+			ResourcePattern: "io.modelcontextprotocol.anonymous/*",
+		},
```
EItanya: What's the difference between `run` and `deploy`?

internal/registry/api/handlers/v0/deployments.go
```diff
@@ -217,14 +236,21 @@ func RegisterDeploymentsEndpoints(api huma.API, basePath string, registry servic
 			return nil, huma.Error400BadRequest("Invalid server name encoding", err)
 		}
 
+		switch input.ResourceType {
+		case "", "mcp", "agent":
+			// Valid resource types
+		default:
+			return nil, huma.Error400BadRequest("Invalid resource type. Must be 'mcp' or 'agent'")
+		}
```
EItanya, line 237: Why does this need to check resource type. The op is `remove-server`

internal/mcp/registryserver/server.go
inFocus7: Yeah, it does through the authN middleware which would set the session w/ claims in context.

But i wonder if using the same authn/z provider makes sense for the mcp server if it's for agents to communicate with, not users. The authn/z would be going through whatever is configured by the provider (e.g. an idp), and I'm not sure what sort of 'identity' + tokens an agent may have 🤔 

> // All endpoints are restricted to published content to keep the surface area safe for unauthenticated agents.

internal/registry/api/handlers/v0/auth/none.go
inFocus7: push -> pushes a resource to registry so it exists there
publish -> making the resource public & ready for use

internal/registry/api/handlers/v0/auth/none.go
inFocus7: read is just seeing existence, pull is being able to pull the resource. although tbf i only noticed that `skills` was the only resource with 'pull', other resource types did not.

i'm going to think about this a bit more to see what makes most sense; i was basing actions on general user actions, but i'm going to reframe in a 'from the pov of registry' way

internal/registry/api/handlers/v0/auth/none.go
inFocus7: run is for running something from registry _locally_ with (based on `arctl run`), deploying is deploying to whatever runtime is chosen (`arctl deploy`/deploying in server).

i think we can condense "pull" and "run" into a singular "pull" since they're very close in purpose, and thinking about it `run` is more of a cli wording, even though it's just pulling/reading from the registry

internal/registry/api/handlers/v0/deployments.go
inFocus7: I'll remove it, it's a leftover from when i had authz checks on this layer 👍🏼

inFocus7: Actually, I'm going to rename this + the db calls.

from db
> // RemoveServer removes a deployment (works for any resource type)

from huma description
> Description: "Remove a resource (MCP server or agent) from deployed state",

It's confusing since everything says "remove server", but it's technically used for general deployment removal. I'm renaming this to `RemoveDeployment`, similar to the other XDeployment naming to avoid confusion 👍🏼

internal/registry/api/handlers/v0/auth/none.go
inFocus7: i simplified actions and removed pull + run

- pull: based on a skill cli cmd, which is actually just reading the artifact (so 'read' action from perspective of registry)
- run: for running from artifact locally (so 'read' again, from perspective or registry)

internal/registry/api/handlers/v0/auth/none.go
EItanya: I think we should get rid of publish, that's basically just editing the resource?

internal/mcp/registryserver/server.go
EItanya: For now let's use the same Auth. This may be used by agents, or by tools like cursor/vscode which are user facing

internal/registry/api/handlers/v0/auth/none.go
inFocus7: technically yeah. although i'm pro-keeping it separate, since they feel like distinct concerns.

e.g. just because i'm allowed to edit a document's metadata, doesn't mean i should be able to make it visible/invisible for others to digest.

EItanya: Maybe admin then? To me an entire layer of RBAC to perform one action, other than delete, seems a bit onerous to teach new users.

inFocus7: Yeah, I think that should be good for now. I was a bit against it since this repo currently doesn't have much of an identity (admin vs. non), but saw that there were some permission checks that does `"resourcePermissions: *" mean admin`, so I can stick with that logic in the repo 🫡 

I was going to remove the Permissions action, but saw it used elsewhere in the code and had testing around it, so I'm not sure if it's safe to. My main concern is its usage in the `/auth` handlers, which I have no context on and don't really know _how_ they're in a real user flow.

internal/registry/database/postgres.go
```diff
@@ -2460,63 +2722,94 @@ func (db *PostgreSQL) UpdateDeploymentConfig(ctx context.Context, tx pgx.Tx, ser
 	query := `
 		UPDATE deployments
 		SET config = $2
-		WHERE server_name = $1
+		WHERE server_name = $1 AND version = $2 AND resource_type = $3
```
ilackarms, line 2866: i think you need to update this to `SET config = $4` for this to work

internal/registry/database/postgres.go
```diff
@@ -2460,63 +2722,94 @@ func (db *PostgreSQL) UpdateDeploymentConfig(ctx context.Context, tx pgx.Tx, ser
...
+		Type: artifactType,
+	}); err != nil {
+		return err
+	}
+
 	executor := db.getExecutor(tx)
 
 	query := `
 		UPDATE deployments
 		SET status = $3
```
ilackarms: same as above, i.e. `SET status = $4`

internal/registry/database/postgres.go
inFocus7: nice catch, updated (same to other)! 👍🏼

PR #87, opened 2026-01-06
feat: semantic search

ilackarms:


internal/cli/embeddings.go
```diff
@@ -0,0 +1,319 @@
...
+	cfg := config.NewConfig()
+	if !cfg.Embeddings.Enabled {
+		return fmt.Errorf("embeddings are disabled (set AGENT_REGISTRY_EMBEDDINGS_ENABLED=true)")
+	}
+
+	if cfg.Embeddings.Dimensions <= 0 {
+		return fmt.Errorf("invalid embeddings dimensions: %d", cfg.Embeddings.Dimensions)
+	}
+
+	db, err := database.NewPostgreSQL(ctx, cfg.DatabaseURL)
```
peterj, line 64: I think this logic should be on the server instead (registry service) and the CLI should just call the functionality - this will allow us to manage the functionality through the UI too

internal/cli/embeddings.go
```diff
@@ -0,0 +1,319 @@
...
+	opts := backfillOptions{
+		limit:      limit,
+		force:      embeddingsForceUpdate,
+		dryRun:     embeddingsDryRun,
+		dimensions: cfg.Embeddings.Dimensions,
+	}
+
+	processServers := embeddingsIncludeServers
+	processAgents := embeddingsIncludeAgents
+	if !processServers && !processAgents {
```
peterj: can this happen? we default to "true" for both of those flags

internal/daemon/docker-compose.yml
```diff
@@ -1,6 +1,6 @@
 services:
   postgres:
-    image: postgres:16-alpine
+    image: pgvector/pgvector:pg16
```
peterj, line 3: I am assuming this is the same as postgre,s but with the vector stuff added?

internal/cli/embeddings.go
EItanya: 100%, that way this can also be managed from the UI

EItanya: In face, the CLI code should be moved to a folder where it cannot import any DB logic so this cannot happen again

internal/models/server_response.go
```diff
@@ -0,0 +1,34 @@
...
+// ServerSemanticMeta carries semantic search metadata for servers.
+type ServerSemanticMeta struct {
+	Score float64 `json:"score"`
+}
+
+// ServerResponseMeta mirrors the MCP ResponseMeta but adds semantic metadata.
+type ServerResponseMeta struct {
+	Official *apiv0.RegistryExtensions `json:"io.modelcontextprotocol.registry/official,omitempty"`
+	Semantic *ServerSemanticMeta       `json:"agentregistry.solo.io/semantic,omitempty"`
+}
```
EItanya, line 16: I believe this change means that our server is no longer compliant with the spec, is that something we care about? @pmuir

internal/registry/database/testutil.go
```diff
@@ -19,40 +19,66 @@ const templateDBName = "agent_registry_test_template"
 // ensureTemplateDB creates a template database with migrations applied
 // Multiple processes may call this, so we handle race conditions
 func ensureTemplateDB(ctx context.Context, adminConn *pgx.Conn) error {
+	// Serialize template creation/migration across concurrent test processes to
+	// avoid racing on extension creation (which can violate pg_extension_name_index).
+	// Use a global advisory lock key to coordinate.
+	const lockKey int64 = 0x61726567 // "areg" prefix
```
EItanya, line 25: Why did this become necessary, this seems pretty hacky

internal/registry/database/testutil.go
```diff
@@ -19,40 +19,66 @@ const templateDBName = "agent_registry_test_template"
...
+}
+
+func ensureVectorExtension(ctx context.Context, uri string) error {
+	conn, err := pgx.Connect(ctx, uri)
+	if err != nil {
+		return fmt.Errorf("failed to connect to template database for extension install: %w", err)
+	}
+	defer func() { _ = conn.Close(ctx) }()
+
+	if _, err := conn.Exec(ctx, "CREATE EXTENSION IF NOT EXISTS vector"); err != nil {
```
EItanya, line 79: Isn't this done by one of the migrations?

internal/registry/service/registry_service.go
```diff
@@ -961,3 +992,53 @@ func convertServerSpecToServerJSON(spec *types.ServerSpec) *apiv0.ServerJSON {
 		Meta:  nil,
 	}
 }
+
+func (s *registryServiceImpl) prepareSemanticOptions(ctx context.Context, filter interface{}) error {
```
EItanya: This seems like a weird helper function if 90% of the code isn't shared, also checking `filter==nil` will never work because the value being passed in isn't an `interface{}` but a real type.

EItanya (changes requested):

internal/models/server_response.go
ilackarms: no, we should still be complaint with the spec. the published api will look like:

```json
{
  "server": { ... publisher metadata only ... },
  "_meta": {
    "io.modelcontextprotocol.registry/official": {...},
    "agentregistry.solo.io/semantic": {"score": ...}
  }
}
```

where the old api would be:

```json
{
  "server": { ... publisher metadata only ... },
  "_meta": {
    "io.modelcontextprotocol.registry/official": {...}
  }
}
```

i don't beleive the api spec has constraints on the content of the `_meta` field unless i am mistaken?

internal/registry/database/testutil.go
ilackarms: this is necessary because we need to initialize the pg vector extension in our testbed which can lead to races. this is a standard way of doing a global db lock which prevents race conditions (which we were hitting in CI before adding this lock). can you explain what's hacky about it?

internal/registry/database/testutil.go
ilackarms: the tests do not run the migrations on every run (would be slow & prevent parallelization). so we simply clone the db template and run this query on each test, which is why it is needed

internal/registry/service/registry_service.go
ilackarms: the purpose of the helper is to avoid duplicating the same ensureSemanticEmbedding call for both ServerFilter and AgentFilter. so i'm not sure what's weird about it. can you explain?

but yes i will go ahead and drop the outer `filter==nil` check

EItanya: But that's what I'm saying, the caller could just call `s.ensureSemanticEmbedding(ctx, f.Semantic)`? Or is the purpose just the nil check? Maybe we should create an interface for this type? I Don't feel super strongly about this, but in general I shy away from `any/interface{}` types in code, plus this helper function basically just checks `nil`. Maybe we should put this logic on `database.ServerFilter` or something

ilackarms: im fine either way. removed the code as you requested

PR #95, opened 2026-01-20
Implement Kubernetes Runtime for Agent and MCP Deployments

supreme-gg-gg:
High level workflow

AI get_age Proxy Pipeline-2026-01-20-231022

Note that for Agents with MCPs specified in `agent.yaml`, we will resolve them, translate them, save the data into a ConfigMap then mount as `mcp-servers.json` so the agent can connect to them at runtime.

## Usage

To use this (alternatively via UI use the dropdown)

```
arctl agent/mcp build . --push
arctl agent/mcp publish .
arctl agent/mcp deploy <name> --runtime kubernetes --namespace kagent
```

Check your K8s resources / use Kagent dashboard to view and interact with the resources.

Some use cases I tested: deploying agent, MCP servers, agent with MCP server in spec

## Limitations

- Deploying using the UI will might not work because you need to set environment variables (like API KEYs) that are **only** parsed by the CLI, this was the original behaviour and is not addressed by the PR (you can patch the agents manually though)
- When using on MacOS using Docker Desktop and a local cluster (kind / Minikube) you will need to change the `.kube/config` file to use `host.docker.internal` instead of the localhost URL

### DIFF
--- a/internal/runtime/agentregistry_runtime.go
+++ b/internal/runtime/agentregistry_runtime.go
@@ -13,15 +13,14 @@ import (
 	"github.com/agentregistry-dev/agentregistry/internal/runtime/translation/api"
 	"github.com/agentregistry-dev/agentregistry/internal/runtime/translation/kagent"
 	"github.com/agentregistry-dev/agentregistry/internal/runtime/translation/registry"
-
 	v1alpha2 "github.com/kagent-dev/kagent/go/api/v1alpha2"
 	kmcpv1alpha1 "github.com/kagent-dev/kmcp/api/v1alpha1"
 	"go.yaml.in/yaml/v3"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
-	"k8s.io/client-go/tools/clientcmd"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
 )
 
 // fieldManager identifies agentregistry as the field owner for server-side apply.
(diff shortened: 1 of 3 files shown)


internal/models/deployment.go
```diff
@@ -12,4 +12,5 @@ type Deployment struct {
 	Config       map[string]string `json:"config"`
 	PreferRemote bool              `json:"preferRemote"`
 	ResourceType string            `json:"resourceType"` // "mcp" or "agent"
+	Runtime      string            `json:"runtime"`      // "local" or "kubernetes"
```
EItanya, line 15: Nit: in future we should use golang "enums" for this, can make it more clear for the user

internal/runtime/translation/kagent/kagent_translator.go
```diff
@@ -0,0 +1,351 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type translator struct {
+	defaultNamespace string
+}
+
+// Hardcoded default namespace for Kagent Agent CRs
+const DefaultNamespace = "default"
```
EItanya: nit: kagent?

internal/runtime/agentregistry_runtime.go
```diff
@@ -11,11 +11,86 @@ import (
...
+
+func init() {
+	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
+	utilruntime.Must(v1alpha2.AddToScheme(scheme))
+	utilruntime.Must(kmcpv1alpha1.AddToScheme(scheme))
+}
+
+// newClient creates a controller-runtime client with the kagent scheme.
+func newClient(verbose bool) (client.Client, error) {
+	kubeconfigPath := os.Getenv("KUBECONFIG")
```
EItanya: This package has an easier way to do this: https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/client/config/config.go

EItanya (commented):
Looks good overall, just some nits

internal/runtime/translation/kagent/kagent_translator.go
supreme-gg-gg: Sure will update, it's customizable with CLI tho

PR #96, opened 2026-01-22
Discover deployed agent and MCP on Kubernetes / Kagent

supreme-gg-gg:
Extends the `GetDeployments` endpoint to take in a runtime filter like resource type filter. When requested for `runtime == "kubernetes"`, fetch the resources in k8s and return them as well; updates agent registry UI. Uses a label `agentregistry.io/managed` to determine if it's managed by AR or "external" Kagent resources, indicated in the `isExternal` field in the `Deployments` model. Alternatively, we can check against database to determine this.

Screenshot 2026-01-22 at 3 43 57 PM


internal/registry/service/service.go
```diff
@@ -95,5 +95,8 @@ type RegistryService interface {
 	// RemoveServer removes a deployment (works for any resource type)
 	RemoveServer(ctx context.Context, resourceName string, version string) error
 
+	// ListKubernetesDeployments lists all agents and MCP servers from Kubernetes, marking external ones
+	ListKubernetesDeployments(ctx context.Context, namespace string) ([]models.KubernetesResource, error)
```
EItanya: Instead of a new method, can we add an optional filter to `GetDeployments`?

internal/runtime/translation/kagent/kagent_translator.go
```diff
@@ -153,6 +153,10 @@ func (t *translator) translateAgent(agent *api.Agent) (*v1alpha2.Agent, error) {
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      AgentResourceName(agent.Name, agent.Version),
 			Namespace: namespace,
+			// Add a label to identify this resource as managed by agentregistry
+			Labels: map[string]string{
+				"agentregistry.io/managed": "true",
```
peterj: nit: should this be "aregistry.ai" instead? (since we don't own the agentregistry.io domain)

peterj: or I guess `agentregistry.solo.io` perhaps? (since that's what we've been using for metadata too)

supreme-gg-gg: Makes sense I've updated to `agentregistry.solo.io`

PR #100, opened 2026-01-23
various fixes

peterj:
- adding more logging
- if the MCP server has an OCI image, we don't need to rebuild the image (i.e. we don't need a Dockerfile etc.)
- support setting the environment variables for MCP servers
- set the cmd and args for MCP server -- this is used when deploying to k8s, so we know which cmd to call
- add a check for localhost:12121 before starting the daemon, so we can run the server outside of docker (faster dev)

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -91,10 +91,19 @@ func (d *DefaultDaemonManager) IsRunning() bool {
 // isServerResponding checks if the server is responding on port 12121
 func isServerResponding() bool {
 	client := &http.Client{Timeout: 2 * time.Second}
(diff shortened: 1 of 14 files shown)


EItanya (approved):

pkg/daemon/daemon.go
```diff
@@ -70,6 +72,11 @@ func (d *DefaultDaemonManager) Start() error {
 }
 
 func (d *DefaultDaemonManager) IsRunning() bool {
+	// First check if a server is responding on the API port (local or Docker)
+	if isServerResponding() {
+		return true
+	}
```
EItanya, line 78: We should add retries here

PR #109, opened 2026-01-27
Patch kubeconfig for MacOS using local K8s clusters

supreme-gg-gg:
In development setup when your cluster API url is localhost (e.g. Kind, Minikube) and you're on MacOS, we need to patch the kubeconfig file so that it uses `host.docker.internal` to properly reach the API server on host machine. Also needs to disable TLS since the cert is not valid for `host.docker.internal` (it's issued for localhost). Currently this is done manually. This PR automates this process. 

When the CLI launches the daemon, it first checks whether the user is on MacOS & using local cluste and if so patch the values by creating a tmp kuebconfig and mounting that instead in `docker-compose.yml`. This will avoid modifying the kubeconfig on the host.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -7,7 +7,6 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"regexp"
 	"runtime"
 	"strings"
 	"time"
@@ -15,6 +14,7 @@ import (
 	"github.com/agentregistry-dev/agentregistry/internal/daemon"
 	"github.com/agentregistry-dev/agentregistry/internal/version"
 	"github.com/agentregistry-dev/agentregistry/pkg/types"
+	"gopkg.in/yaml.v3"
 )
 
 // DefaultConfig returns the default configuration for the daemon (AgentRegistry OSS daemon)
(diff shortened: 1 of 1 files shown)


pkg/daemon/daemon.go
```diff
@@ -55,11 +58,61 @@ func NewDaemonManager(config *types.DaemonConfig) *DefaultDaemonManager {
...
+		return d.config.ComposeYAML
+	}
+
+	// Patch localhost/127.0.0.1 to host.docker.internal for Docker Desktop on macOS
+	patched := strings.ReplaceAll(string(content), "localhost", "host.docker.internal")
+	patched = strings.ReplaceAll(patched, "127.0.0.1", "host.docker.internal")
+
+	// Replace certificate-authority-data with insecure-skip-tls-verify
+	caRegex := regexp.MustCompile(`(?m)^\s*certificate-authority-data:.*$`)
+	patched = caRegex.ReplaceAllString(patched, "    insecure-skip-tls-verify: true")
```
EItanya: This is a bit too dangerous, we only really want to do this for the `localhost` entries. There may be other entries where this is a bad idea.

supreme-gg-gg: That's valid it should not be replace all, I'll only replace clusters where it's localhost

PR #130, opened 2026-02-03
always re-build when running (add a --no-build flag to skip building)

### LINKED ISSUE #127: arctl agent run should automatically rebuild the agent
we should build by default before run, and perhaps add a flag to not build (e.g. `--no-build`)

peterj:
fixes #127

### DIFF
--- a/internal/cli/agent/run.go
+++ b/internal/cli/agent/run.go
@@ -37,10 +37,10 @@ launches the same chat interface.`,
   arctl agent run dice`,
 }
 
-var noBuild bool
+var buildFlag bool
 
 func init() {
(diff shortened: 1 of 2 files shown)


internal/cli/agent/run.go
```diff
@@ -37,6 +37,12 @@ launches the same chat interface.`,
   arctl agent run dice`,
 }
 
+var noBuild bool
+
+func init() {
+	RunCmd.Flags().BoolVar(&noBuild, "no-build", false, "Skip building the agent before running")
```
EItanya: Can we make this `--build`? may be a bit clearer

peterj: build and set to true by default

PR #134, opened 2026-02-04
add the ability for db factory to create the database too

peterj:


inFocus7 (commented):
What's the use-case? 👀

(from my understanding)
At the moment, this url is used to set up the client/pools for the database, then the db here is used in [the api's calls](https://github.com/agentregistry-dev/agentregistry/blob/1a2d766e3dacb9564657ea82cb0755d8ea07765c/internal/registry/service/registry_service.go#L64). Just wondering how an implementation using "noop" would look like if no connection is initialized (maybe it's a knowledge gap on my end and initial pool isn't needed).

internal/registry/registry_app.go
```diff
@@ -66,21 +66,31 @@ func App(_ context.Context, opts ...types.AppOptions) error {
...
-	if err != nil {
-		return fmt.Errorf("failed to connect to PostgreSQL: %w", err)
-	}
-
-	// Allow implementors to wrap the database, and run additional migrations
-	var db database.Database = baseDB
-	if options.DatabaseFactory != nil {
-		db, err = options.DatabaseFactory(ctx, cfg.DatabaseURL, baseDB, authz)
+	var db database.Database
+	if cfg.DatabaseURL == "noop" && options.DatabaseFactory != nil {
```
inFocus7: IMO we should either do a `cfg.DisableDatabase` approach or rewrite this conditional.

From what i can tell, when deploying based off this repo with `DatabaseURL: "noop"`, then:
1. `options.DatabaseFactory` _will_ be `nil`
2. This will fall through to the `else` statement
3. The baseDB will be created with "noop" url, which I believe would cause errors from the db connection.

If we don't use a new `bool` config, maybe just updating this code to something like below is clearer?:
```go
var db database.Database
if cfg.DatabaseURL == "noop" {
  // noop database without factory would not create a database
  if options.DatabaseFactory != nil {
    ...
  }
} else {
  // existing logic you have
}
```

let me know what you think 👍🏼

internal/registry/registry_app.go
```diff
@@ -66,21 +66,31 @@ func App(_ context.Context, opts ...types.AppOptions) error {
...
-	baseDB, err := internaldb.NewPostgreSQL(ctx, cfg.DatabaseURL, authz)
-	if err != nil {
-		return fmt.Errorf("failed to connect to PostgreSQL: %w", err)
-	}
-
-	// Allow implementors to wrap the database, and run additional migrations
-	var db database.Database = baseDB
-	if options.DatabaseFactory != nil {
-		db, err = options.DatabaseFactory(ctx, cfg.DatabaseURL, baseDB, authz)
+	var db database.Database
```
inFocus7, line 73: can we add a comment here or on the description with the context of the changes so it's clearer for developers to know when to use (or not use) "noop"?

inFocus7 (approved):

PR #136, opened 2026-02-08
fix: hide registry flags from mcp init help output

### LINKED ISSUE #115: arctl mcp init contains flags it shouldn't
`arctl mcp init python --help`

Notice the in the global flags, theres:

```
  --registry-token string   Registry bearer token (overrides ARCTL_API_TOKEN)
      --registry-url string     Registry base URL (overrides ARCTL_API_BASE_URL; default http://localhost:12121)
```

These two flags shouldn't be there for the mcp init command as we don't do anything with the registry in that command.

joelklabo:
## Summary

Fixes #115 — `arctl mcp init` no longer shows `--registry-token` and `--registry-url` in its help output.

**Problem:** The root command defines `--registry-url` and `--registry-token` as `PersistentFlags`, which Cobra inherits to all subcommands including `mcp init`. Since `mcp init` is a local scaffolding command that never connects to a registry, these flags are confusing.

**Fix:**
1. Override `PersistentPreRunE` on `InitCmd` with a no-op to prevent the root's registry-connection logic from running
2. Shadow the root's registry flags with hidden local persistent flags on `InitCmd` so they don't appear in help output

**Before:**
```
Global Flags:
      --registry-token string   Registry bearer token (overrides ARCTL_API_TOKEN)
      --registry-url string     Registry base URL (overrides ARCTL_API_BASE_URL; default http://localhost:12121)
  -v, --verbose                 Enable verbose output
```

**After:**
```
Global Flags:
  -v, --verbose   Enable verbose output
```

This applies to `mcp init`, `mcp init go`, and `mcp init python`.

## Test plan
- [x] `arctl mcp init --help` no longer shows registry flags
- [x] `arctl mcp init go --help` no longer shows registry flags
- [x] `arctl mcp init python --help` no longer shows registry flags
- [x] `arctl mcp list --help` still shows registry flags correctly
- [x] Build succeeds

### DIFF
--- a/internal/cli/mcp/init.go
+++ b/internal/cli/mcp/init.go
@@ -23,6 +23,11 @@ var InitCmd = &cobra.Command{
 This command provides subcommands to initialize a new MCP server project
 using one of the supported frameworks.`,
 	RunE: runInit,
+	// Override root's PersistentPreRunE — mcp init is a local scaffolding
+	// command that does not need a registry connection.
+	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+		return nil
+	},
 }
 
 var (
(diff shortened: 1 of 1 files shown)


internal/cli/mcp/init.go
```diff
@@ -23,6 +23,11 @@ var InitCmd = &cobra.Command{
 This command provides subcommands to initialize a new MCP server project
 using one of the supported frameworks.`,
 	RunE: runInit,
+	// Override root's PersistentPreRunE — mcp init is a local scaffolding
+	// command that does not need a registry connection.
+	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
```
peterj, line 28: I think there's the same in the agent init command, so we should perhaps implement the same there

Issue #77, opened 2025-12-10
add a unit test that ensures all commands are in the cli


peterj:

timflannagan:
@groundnuty Can you comment on this issue so we can assign you for bookkeeping purposes?

PR #137, opened 2026-02-08
Add test ensuring all CLI commands are registered

joelklabo:
## Summary

Addresses #77. Adds `TestAllCommandsRegistered` in `pkg/cli/root_test.go` which verifies that every expected top-level command and subcommand is present in the cobra command tree.

- Checks all 8 top-level commands (agent, configure, embeddings, export, import, mcp, skill, version)
- Checks subcommands for agent (12), mcp (11), skill (8), and embeddings (1)
- If a new command is added to the source but not to the expected list (or vice versa), the test fails with a clear diff

## Test plan

- [x] `go test ./pkg/cli/ -run TestAllCommandsRegistered -v` passes
- [x] All existing tests still pass (`go test ./pkg/cli/... ./internal/cli/...`)
- [x] `go vet ./pkg/cli/` clean

### DIFF
--- a/pkg/cli/root_test.go
+++ b/pkg/cli/root_test.go
@@ -1,0 +1,124 @@
+package cli
+
+import (
+	"sort"
+	"testing"
+
+	"github.com/spf13/cobra"
+)
+
+// TestAllCommandsRegistered ensures every expected CLI command and subcommand
+// is registered on the root cobra command tree. If a new command is added to
+// the source but not to the expected list (or vice versa), this test fails.
+func TestAllCommandsRegistered(t *testing.T) {
+	root := Root()
+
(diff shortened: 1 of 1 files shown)


joelklabo:
The lint is failing because the code uses the `sort` package but the project requires using `slices` instead (per the `forbid-pkg-errors` depguard rule).\n\n**Fix needed:** Replace:\n- `import "sort"` → `import "slices"`\n- `sort.Strings(names)` → `slices.Sort(names)`\n- `sort.Strings(sortedExpected)` → `slices.Sort(sortedExpected)`\n- `sort.Strings(sortedGot)` → `slices.Sort(sortedGot)`\n\nWould you like me to push this fix to the branch?

PR #138, opened 2026-02-08
Randomize RuntimeDir to prevent concurrent run collisions

### LINKED ISSUE #126: ensure the folder where docker compose and agent config are stored is random on each run
we should probably have this folder be generated randomly -- we default to some base/root and then append random string to the name

```
RuntimeDir         string `env:"RUNTIME_DIR" envDefault:"/tmp/arctl-runtime"`
```

joelklabo:
# Description

Fixes #126. Appends a random hex suffix to the default `RuntimeDir` (`/tmp/arctl-runtime`) so that concurrent `arctl` sessions each get their own docker compose and config directory.

- When `AGENT_REGISTRY_RUNTIME_DIR` is **not set**: default becomes `/tmp/arctl-runtime-<16 hex chars>` (unique per invocation)
- When `AGENT_REGISTRY_RUNTIME_DIR` **is set**: the user's explicit value is used as-is (no suffix)
- Added 3 tests in `config_test.go`: suffix format, uniqueness between calls, and env var override

# Change Type

/kind cleanup

```release-note
NONE
```

## Test plan

- [x] `go test ./internal/registry/config/ -v` — 3 new tests pass
- [x] `go build ./...` — full project builds cleanly
- [x] `go vet ./internal/registry/config/` — clean

### DIFF
--- a/internal/registry/config/config.go
+++ b/internal/registry/config/config.go
@@ -1,7 +1,10 @@
 package config
 
 import (
+	"crypto/rand"
+	"encoding/hex"
 	"log"
+	"os"
 
 	env "github.com/caarlos0/env/v11"
 	"github.com/joho/godotenv"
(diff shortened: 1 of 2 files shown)


peterj (approved):

peterj:
closes https://github.com/agentregistry-dev/agentregistry/issues/126

PR #142, opened 2026-02-08
fix: skip daemon for offline commands (agent init, mcp init)

### LINKED ISSUE #31: agent init shouldn't depend on the registry running

joelklabo:
## Summary

Fixes #31.

- `agent init` and `mcp init` (including subcommands like `mcp init go`, `mcp init python`) only do local file I/O (scaffolding templates, creating directories) and never call the registry API
- The `PersistentPreRunE` hook forced daemon startup and API client creation for **all** commands, causing `agent init` to fail when the daemon was not running:
  ```
  Error: API client not initialized: failed to reach API at http://localhost:12121/v0
  ```
- Added `isOfflineCommand()` check that short-circuits `PersistentPreRunE` for commands that don't need network access
- New offline commands can be added by appending to the `offlineCommands` slice

## Changes

- `pkg/cli/root.go`: Added `offlineCommands` list and `isOfflineCommand()` function; early return in `PersistentPreRunE` for matching commands
- `pkg/cli/root_test.go`: 8 test cases covering offline and online commands, including subcommand matching

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./...` passes (all existing + 8 new tests)
- [x] `arctl agent init adk python dice` works without daemon running
- [x] Online commands (`agent list`, `mcp list`, etc.) still require daemon as before

### DIFF
--- a/pkg/cli/root.go
+++ b/pkg/cli/root.go
@@ -46,6 +46,11 @@ var rootCmd = &cobra.Command{
 	Short: "Agent Registry CLI",
 	Long:  `arctl is a CLI tool for managing agents, MCP servers and skills.`,
 	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+		// Commands that work entirely offline don't need the daemon or API client.
+		if isOfflineCommand(cmd) {
+			return nil
+		}
+
 		baseURL, token := resolveRegistryTarget()
 
 		dm := cliOptions.DaemonManager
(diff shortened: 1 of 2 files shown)


peterj:
fixed by https://github.com/agentregistry-dev/agentregistry/pull/189

PR #143, opened 2026-02-08
Fix jq parse error in get-arctl download script

### LINKED ISSUE #45: Getting a jq error in the download script
```
curl -fsSL https://raw.githubusercontent.com/agentregistry-dev/agentregistry/main/scripts/get-arctl | bash
jq: parse error: Invalid numeric literal at line 1, column 9
```

joelklabo:
# Description

Fixes #45 — The `get-arctl` install script fails with `jq: parse error: Invalid numeric literal at line 1, column 9` during the installed version check.

### Root cause

`checkArctlInstalledVersion()` pipes `arctl version` output to `jq -r '.arctl_version'`, expecting JSON. But `arctl version` outputs plain text:

```
arctl version v0.1.1
Git commit: abc123
Build date: 2026-01-15
```

# Change Type

```
/kind fix
```

# Changelog

```release-note
NONE
```

### Changes

1. **Line 129**: Parse version from text output using `awk '{print $NF}'` instead of `jq`
2. **Line 115**: Suppress jq stderr (`2>/dev/null`) so GitHub API errors produce a clean failure message instead of a confusing jq parse error

### Testing

- Verified `arctl version` output format matches the awk extraction pattern
- The existing flow (version comparison, download, install) is unchanged

### DIFF
--- a/internal/cli/version.go
+++ b/internal/cli/version.go
@@ -41,19 +41,36 @@ var VersionCmd = &cobra.Command{
 		}
 
 		serverVersion, err := apiClient.GetVersion()
-		if err == nil {
-			output.ServerVersion = serverVersion.Version
-			output.ServerGitCommit = serverVersion.GitCommit
-			output.ServerBuildDate = serverVersion.BuildTime
-
-			if semver.IsValid(version.EnsureVPrefix(serverVersion.Version)) && semver.IsValid(version.EnsureVPrefix(version.Version)) {
-				compare := semver.Compare(version.EnsureVPrefix(version.Version), version.EnsureVPrefix(serverVersion.Version))
-				switch compare {
-				case 1:
-					output.UpdateRecommendation = "CLI version is newer than server version. Consider updating the server."
-				case -1:
-					output.UpdateRecommendation = "Server version is newer than CLI version. Consider updating the CLI."
+		if err != nil {
+			// If server version fetch fails, we still output arctl version info
+			if jsonOutput {
+				jsonBytes, err := json.MarshalIndent(output, "", "  ")
+				if err != nil {
+					fmt.Printf("Error marshaling JSON: %v\n", err)
+					return
 				}
+				fmt.Println(string(jsonBytes))
+				return
+			}
+			// Human-readable output
+			fmt.Printf("arctl version %s\n", output.ArctlVersion)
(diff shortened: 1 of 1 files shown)


scripts/get-arctl
```diff
@@ -126,7 +126,7 @@ checkDesiredVersion() {
 # if it needs to be changed.
 checkArctlInstalledVersion() {
   if [[ -f "${ARCTL_INSTALL_DIR}/${BINARY_NAME}" ]]; then
-    local version=$("${ARCTL_INSTALL_DIR}/${BINARY_NAME}" version | jq -r '.arctl_version')
+    local version=$("${ARCTL_INSTALL_DIR}/${BINARY_NAME}" version 2>/dev/null | head -1 | awk '{print $NF}')
```
EItanya: Can we add an option to `arctl` for version to output JSON if it isn't already there?

joelklabo:
Added `--json` flag to `arctl version` command. The get-arctl script now uses `arctl version --json | jq -r '.arctl_version'` for reliable version parsing. This avoids fragile text parsing and makes the script more robust.

internal/cli/version.go
```diff
@@ -16,39 +17,76 @@ func SetAPIClient(client *client.Client) {
...
+			ArctlVersion: version.Version,
+			GitCommit:    version.GitCommit,
+			BuildDate:    version.BuildDate,
+		}
+
 		serverVersion, err := apiClient.GetVersion()
-		if err != nil {
-			fmt.Printf("Error getting server version: %v\n", err)
-			return
+		if err == nil {
```
peterj: can you flip the condition here and check for error then return?

joelklabo:
@peterj pushed a fix (79ec9ab) that flips the error check condition as requested. Now checks for error first and returns early, making the success path more linear. Ready for re-review.

joelklabo:
CI checks don't appear to have run on the latest commit (79ec9ab). Could you please re-run CI or re-review? Happy to address any additional feedback.

peterj:
Fixes #45

peterj:
closing in favor of https://github.com/agentregistry-dev/agentregistry/pull/262 (it was easier to create a new PR, still crediting you @joelklabo  - thanks again!)

PR #153, opened 2026-02-09
make sure agent and mcp and other resources always go to the same nam…

### LINKED ISSUE #147: when deploying to k8s, the mcp server goes to 'kagent' while agent goes to 'default'
if you don't provide the namespace when deploying, the agent will go to default namespace and the mcp server to kagent. we need to go to default (if no namespace is provided) or both resources to the same namespace the user provided .

peterj:
…espace

fixes #147

### DIFF
--- a/internal/cli/agent/deploy.go
+++ b/internal/cli/agent/deploy.go
@@ -134,7 +134,11 @@ func deployKubernetes(name, version string, config map[string]string, namespace 
 		return fmt.Errorf("failed to deploy agent: %w", err)
 	}
 
-	fmt.Printf("Agent '%s' version '%s' deployed to kubernetes runtime in namespace '%s'\n", deployment.ServerName, deployment.Version, namespace)
+	ns := namespace
+	if ns == "" {
+		ns = "(default)"
+	}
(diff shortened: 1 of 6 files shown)


internal/cli/agent/deploy.go
```diff
@@ -142,5 +142,5 @@ func init() {
 	DeployCmd.Flags().String("version", "latest", "Agent version to deploy")
 	DeployCmd.Flags().String("runtime", "local", "Deployment runtime target (local, kubernetes)")
 	DeployCmd.Flags().Bool("prefer-remote", false, "Prefer using a remote source when available")
-	DeployCmd.Flags().String("namespace", "", "Kubernetes namespace for agent deployment")
+	DeployCmd.Flags().String("namespace", "default", "Kubernetes namespace for agent deployment")
```
EItanya: Can't we just point to the `const` variable defined in the other package?

internal/runtime/translation/kagent/kagent_translator.go
```diff
@@ -18,7 +18,7 @@ type translator struct {
 	defaultNamespace string
 }
 
-const DefaultNamespace = "kagent"
+const DefaultNamespace = "default"
```
EItanya: Why does this make a difference if we just use this var from the flag? In general most tools actually just set this to `""` and use the default namespace in the context unless specified, what do you think?

PR #165, opened 2026-02-13
clean up the publish command for mcp/agents

peterj:
this PR does a bit of a clean up and refactoring, especially around the publish commands.

Notable changes: 
-  `mcp publish` command doesn't build, you must call `mcp build` OR build the image yourself
- the publish now takes the `--type` flag which can be oci/npm/pypi -- this allows publishing existing mcp servers as well as publishing a server that you scaffolded by running the command in the folder with mcp.yaml
- the remove command is gone; the delete command removes the entry from the registry and the reconciliation should make sure to delete any deployments of that entry


internal/cli/common/common.go
```diff
@@ -0,0 +1,116 @@
...
+	if flagVersion != "" {
+		return flagVersion
+	}
+	if manifestVersion != "" {
+		return manifestVersion
+	}
+	return "latest"
+}
+
+// agentNameRegex matches the database constraint for agent names:
```
krisztianfekete: Do we want to do the same for MCP as well?

peterj: refactored that part as well -- so we have pkg/validators now for naming validations. there's still work that could be done there in terms of unifying the naming (e.g. for mcp servers we follow a certain naming format (e.g. user/server), while for agents we don't).

EItanya (approved):

ilackarms (approved):

PR #169, opened 2026-02-16
Makefile: Simplify environment bootstrapping

timflannagan:
Introduce the "run" and "down" Makefile targets that provide a single target I can use for both local & CI. Updates the build.yml workflow to adopt that new target, which removes some of the existing spaghetti as well.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,22 +38,14 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-node-
 
-    - name: Download Go dependencies
-      run: go mod download
-
     - name: Build UI
       run: make build-ui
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -38,22 +38,14 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-node-
 
-    - name: Download Go dependencies
-      run: go mod download
```
timflannagan, line 42: Hmm not sure this is correct. Lemme double check what's right here.

timflannagan: Oh nvm, had it right the first time. The build-cli command calls this already so there were three instances where go mod download was being called before #167 was merged earlier today lol.

PR #171, opened 2026-02-16
Makefile: Consolidate lint targets and remove redundant fmt

timflannagan:
Removes the separate lint-fix and fmt targets. The lint target now accepts GOLANGCI_LINT_ARGS for flexibility:

```bash
make lint
make lint GOLANGCI_LINT_ARGS=--fix
```

This reduces target sprawl while maintaining the same functionality. golangci-lint already includes gofmt and goimports, making the standalone fmt target redundant and it's unused in CI.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ LDFLAGS := \
 # Local architecture detection to build for the current platform
 LOCALARCH ?= $(shell uname -m | sed 's/x86_64/amd64/' | sed 's/aarch64/arm64/')
 
-.PHONY: help install-ui build-ui clean-ui build-cli build install dev-ui run down test test-integration test-coverage test-coverage-report clean fmt lint all release-cli docker-compose-up docker-compose-down docker-compose-logs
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -246,9 +240,11 @@ release-cli: bin/arctl-darwin-amd64.sha256
 release-cli: bin/arctl-darwin-arm64.sha256  
 release-cli: bin/arctl-windows-amd64.exe.sha256
 
+GOLANGCI_LINT_ARGS ?=
+
 .PHONY: lint
-lint: golangci-lint ## Run golangci-lint linter
-	$(GOLANGCI_LINT) run
+lint: golangci-lint ## Run golangci-lint linter (use GOLANGCI_LINT_ARGS=--fix to auto-fix)
```
timflannagan, line 246: On the fence about the default here. Could go either way.

Makefile
```diff
@@ -246,9 +240,11 @@ release-cli: bin/arctl-darwin-amd64.sha256
...
+	$(GOLANGCI_LINT) run $(GOLANGCI_LINT_ARGS)
```
peterj, line 247: I'd just add --fix by default

timflannagan: I'll throw this on my list to fix at some point. Agreed with it being the default.

PR #180, opened 2026-02-18
Makefile: Misc. updates, register golangci-lint as a tool, etc

timflannagan:
Several misc. updates to the Makefile. Continuation of several PRs that landed recently. The biggest change is registering golangci-lint as a Go tool dependency, which is consistent with how it's managed in other repositories.


peterj (approved):

timflannagan:
created https://github.com/agentregistry-dev/agentregistry/issues/181 for the flake

PR #182, opened 2026-02-18
ci: Add stalebot workflow for PRs and issues

timflannagan:
Adapted from kgateway's stalebot implementation. PRs are marked stale after 14 days with 3 days to respond before closing. Issues use a longer cadence of 60 days stale with 14 days to respond. Both can be exempted with the "no stalebot" label.

### DIFF
--- a/.github/workflows/stalebot.yml
+++ b/.github/workflows/stalebot.yml
@@ -1,0 +1,62 @@
+name: Stalebot
+
+on:
+  workflow_dispatch: {}
+  schedule:
+    # Daily at 6am EST
+    - cron: "0 10 * * *"
+
+env:
+  DAYS_BEFORE_ISSUE_STALE: 60
+  DAYS_BEFORE_ISSUE_CLOSE: 14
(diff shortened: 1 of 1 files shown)


.github/workflows/stalebot.yml
```diff
@@ -0,0 +1,62 @@
...
+  workflow_dispatch: {}
+  schedule:
+    # Daily at 6am EST
+    - cron: "0 10 * * *"
+
+env:
+  DAYS_BEFORE_ISSUE_STALE: 60
+  DAYS_BEFORE_ISSUE_CLOSE: 14
+  DAYS_BEFORE_PR_STALE: 14
+  DAYS_BEFORE_PR_CLOSE: 3
```
timflannagan, line 13: Most of these values were arbitrarily chosen. Tried to be reasonable while aggressive on the values. We could be even more aggressive if we want since it's easier to loosen that over time imo.

peterj: let's go with it -- we can always adapt if it's too aggressive

PR #189, opened 2026-02-19
locking down to specific python and dependency versions

### LINKED ISSUE #186: lock down to specific dependency versions in pyproject.toml
e.g. FastMCP and others. we got broken by the latest fastmcp release because our version selection is too loose.

peterj:
Setting explicit dependency versions for the MCP project and the agent project. Note that the agent project right now requires 3.13, as one of it's dependencies requires is (kagent-skills). We could lower that version later [once this is fixed](https://github.com/kagent-dev/kagent/issues/1329).

The first commit also updates CLI so it doesn't try to connect/launch to the daemon when running the init command.

Resolves #186

### DIFF
--- a/pkg/cli/root.go
+++ b/pkg/cli/root.go
@@ -66,6 +66,46 @@ func Execute() {
 	}
 }
 
+var rootCmd = &cobra.Command{
+	Use:   "arctl",
+	Short: "Agent Registry CLI",
+	Long:  `arctl is a CLI tool for managing agents, MCP servers and skills.`,
+	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+		baseURL, token := resolveRegistryTarget(os.Getenv)
+		skipSetup, autoStartDaemon := preRunBehavior(cmd, baseURL)
+		if skipSetup {
+			return nil
+		}
+
+		c, err := preRunSetup(cmd.Context(), baseURL, token, autoStartDaemon)
+		if err != nil {
+			return err
+		}
+
+		APIClient = c
+		mcp.SetAPIClient(c)
+		agent.SetAPIClient(c)
+		agentutils.SetDefaultRegistryURL(c.BaseURL)
+		skill.SetAPIClient(c)
+		cli.SetAPIClient(c)
+		return nil
+	},
+}
+
+func init() {
+	rootCmd.PersistentFlags().StringVar(&registryURL, "registry-url", os.Getenv("ARCTL_API_BASE_URL"), "Registry base URL (overrides ARCTL_API_BASE_URL; default http://localhost:12121)")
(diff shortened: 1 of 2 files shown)


internal/cli/mcp/frameworks/python/templates/src/core/server.py.tmpl
timflannagan, line 1: Are these changes related to the ADK replace pin?

pkg/cli/root.go
```diff
@@ -167,20 +153,92 @@ func shouldAutoStartDaemon(targetURL string) bool {
...
-	parsed, err = url.Parse("http://" + raw)
 	if err != nil {
-		return nil
+		return nil, fmt.Errorf("API client not initialized: %w", err)
 	}
-	return parsed
+	return c, nil
+}
+
+func applyAPIClientToSubcommands(c *client.Client) {
```
timflannagan: nit: I'd advocate for moving all of this logic until EIF at the top of the file. Easier to read for cobra-based CLIs where you have the rootCmd that acts as a driver + init() commands.

pkg/cli/root.go
```diff
@@ -167,20 +153,92 @@ func shouldAutoStartDaemon(targetURL string) bool {
...
+		if skipSetup {
+			return nil
+		}
+
+		c, err := preRunSetup(cmd.Context(), baseURL, token, autoStartDaemon)
+		if err != nil {
+			return err
+		}
+
+		applyAPIClientToSubcommands(c)
```
timflannagan: Looks like this function is used here & for testing purposes. Can't we inline this instead and skip the unit tests?

pkg/cli/root.go
```diff
@@ -167,20 +153,92 @@ func shouldAutoStartDaemon(targetURL string) bool {
...
+func applyAPIClientToSubcommands(c *client.Client) {
+	APIClient = c
+	mcp.SetAPIClient(c)
+	agent.SetAPIClient(c)
+	agentutils.SetDefaultRegistryURL(c.BaseURL)
+	skill.SetAPIClient(c)
+	cli.SetAPIClient(c)
+}
+
+var rootCmd = &cobra.Command{
```
timflannagan: We can do this in the future, but some of the globals being used throughout this root package can be dramatically reduced via constructors on each sub-command. Obviously doesn't block these changes though.

internal/cli/mcp/frameworks/python/templates/src/core/server.py.tmpl
peterj: the requires here is just to be sure -- we do set the version in .python-version file too, but it's clearer if it's here as well (note we depend on 3.13 because of one o the kagent libraries)

EItanya:
This has been resolved on the kagent side, we are now unit testing down to `3.10`

PR #144, opened 2026-02-08
Implement agent add-skill subcommand

joelklabo:
## Summary

Implements the `arctl agent add-skill` command (fixes #64), replacing the stub with a full implementation. Skills can be added to an agent project from three sources:

- **Docker image**: `arctl agent add-skill my-skill --image docker.io/org/skill:latest`
- **Registry**: `arctl agent add-skill my-skill --registry-skill-name cool-skill`
- **Local scaffold**: `arctl agent add-skill my-skill --scaffold`

### What changed

| File | Change |
|------|--------|
| `pkg/models/manifest.go` | Added `SkillRef` type and `Skills []SkillRef` field to `AgentManifest` |
| `internal/cli/agent/add-skill.go` | Full implementation: flag parsing, mode detection, duplicate name check, manifest save, scaffolding |
| `internal/cli/agent/frameworks/common/manifest_manager.go` | Added skill validation: name required, exactly one of image/path/registrySkillName |
| `internal/cli/agent/add-skill_test.go` | 12 tests covering all three modes, validation rules, duplicate detection, and error cases |

### Design decisions

- Follows the same patterns as `add-mcp`: load manifest → validate → append → save → post-process
- `SkillRef` is separate from the existing `SkillJSON` (registry metadata) — it only stores the reference needed in the agent manifest
- Scaffold mode reuses the existing `skill/templates.Generator` with `Empty: true` for minimal scaffolding
- Validation enforces mutual exclusivity: exactly one source type per skill

### Testing

All 34 tests in `internal/cli/agent/` pass, including 12 new tests:
- `TestAddSkillWithImage`, `TestAddSkillWithRegistry`, `TestAddSkillWithScaffold`
- `TestAddSkillDuplicateName`, `TestAddSkillNoFlags`, `TestAddSkillMultipleFlags`
- 6 `TestSkillValidation` subtests for each validation rule

PR #198, opened 2026-02-20
cli: Add support for agent add-skills command

### LINKED ISSUE #64: implement "add-skill" subcommand for agents
add-skill is executed in the context of an initialized agent (e.g. `arctl agent add-skill`). The user can point to an existing docker image that packages the skill OR we can scaffold an empty skill within the agent.

timflannagan:
# Description

Implements the `arctl agent add-skill` command (issue #64) for adding skills to an agent project.

Supersedes the #144 PR and addresses feedback from Peter. Additionally, removes the option
to locally scaffold a skill since there's overlap with the skills init command. For now, we can
leverage that command instead of maintaining different ways to scaffolding skills locally.

Fixes #64 

# Change Type

/kind feature

# Changelog

```release-note
Add `arctl agent add-skill` command for adding skills to agent projects
```

### DIFF
--- a/pkg/models/manifest.go
+++ b/pkg/models/manifest.go
@@ -24,8 +24,6 @@ type SkillRef struct {
 	Name string `yaml:"name" json:"name"`
 	// Image is a Docker image containing the skill (for image type).
 	Image string `yaml:"image,omitempty" json:"image,omitempty"`
-	// Path is the local directory path for a scaffolded skill (for local type).
-	Path string `yaml:"path,omitempty" json:"path,omitempty"`
 	// RegistryURL is the registry URL for pulling the skill (for registry type).
 	RegistryURL string `yaml:"registryURL,omitempty" json:"registryURL,omitempty"`
 	// RegistrySkillName is the skill name in the registry.
--- a/internal/cli/agent/add-skill_test.go
+++ b/internal/cli/agent/add-skill_test.go
@@ -172,11 +172,7 @@ func TestSkillValidation(t *testing.T) {
 			skills:  []models.SkillRef{{Name: "s1", Image: "img:latest"}},
 			wantErr: false,
 		},
-		{
-			name:    "valid path skill",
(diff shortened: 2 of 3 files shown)


internal/cli/agent/add-skill.go
```diff
@@ -2,20 +2,124 @@ package agent
...
+	name := args[0]
+
+	if err := addSkillCmd(name); err != nil {
+		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+		os.Exit(1)
+	}
+	return nil
+}
+
+func addSkillCmd(name string) error {
```
timflannagan, line 62: Note: this command only sets the metadata on the agent.yaml right now. We'll need some follow-up changes in order to make `arctl agent run ...` leverage added tools (like they do with MCP servers). Dived into this briefly locally, but need to learn more about how kagent handles all of this. There's a couple of options here though.

pkg/models/manifest.go
```diff
@@ -14,9 +14,24 @@ type AgentManifest struct {
 	Version           string          `yaml:"version,omitempty" json:"version,omitempty"`
 	TelemetryEndpoint string          `yaml:"telemetryEndpoint,omitempty" json:"telemetryEndpoint,omitempty"`
 	McpServers        []McpServerType `yaml:"mcpServers,omitempty" json:"mcpServers,omitempty"`
+	Skills            []SkillRef      `yaml:"skills,omitempty" json:"skills,omitempty"`
 	UpdatedAt         time.Time       `yaml:"updatedAt,omitempty" json:"updatedAt,omitempty"`
 }
 
+// SkillRef represents a skill reference in the agent manifest.
+type SkillRef struct {
```
timflannagan, line 22: Removed --path support as well. I think this is still something worth supporting for local dev, but wanted to focus on the image use case first.

timflannagan: Err removed the Path field on this struct. IIRC, we didn't have formal --path support.

PR #201, opened 2026-02-20
ci: Extract lint jobs into standalone workflow

timflannagan:
# Description

Extracts linting from build.yml into a dedicated lint.yml workflow. The previous setup unnecessarily ran a full Next.js production build before linting. Now we have parallel lint-go and lint-ui jobs, with a new lint-ui Makefile target that only installs dependencies.

/kind cleanup

```release-note
NONE
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -46,26 +46,3 @@ jobs:
 
     - name: Run Integration Tests
       run: make test
-
-  lint:
-    runs-on: ubuntu-latest
-    steps:
-    - uses: actions/checkout@v4
-
-    - uses: actions/setup-go@v6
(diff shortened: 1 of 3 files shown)


peterj (approved):

timflannagan:
Ah right, we need to update branch protection rules in order to merge this. @peterj I don't have perms to do that, happy to either do that myself, or hold off until someone can update those rules

peterj:
what needs updating?

timflannagan:
We need to add "Lint/Lint-go" and "Lint/lint-ui" as required jobs and remove "lint" as required. The Verify/verify job also needs to be marked as required.

Issue #204, opened 2026-02-21
refactor: PythonGenerator.Generate mutates caller's AgentConfig via pointer


groundnuty:
## Description

`PythonGenerator.Generate` receives `*common.AgentConfig` and mutates several fields in-place:

```go
// internal/cli/agent/frameworks/adk/python/generator.go

func (g *PythonGenerator) Generate(agentConfig *common.AgentConfig) error {
    agentConfig.Name = validators.PythonSafeName(agentConfig.Name)  // line 36
    // ...
    agentConfig.Framework = "adk"    // line 51
    agentConfig.Language = "python"  // line 52
}
```

The caller in `init.go` currently works correctly because it uses a separate `agentName` string variable for post-`Generate` output. However, any future code that reads `agentConfig.Name` after calling `Generate` would observe the transformed value (`my_agent`) instead of the original (`my-agent`).

This is a latent correctness risk - the function silently changes its input without the caller expecting it.

## Suggested Fix

Work on a local copy inside `Generate` instead of mutating the caller's struct:

```go
func (g *PythonGenerator) Generate(agentConfig *common.AgentConfig) error {
    if agentConfig == nil {
        return fmt.Errorf("agent config is required")
    }

    cfg := *agentConfig
    cfg.Name = validators.PythonSafeName(agentConfig.Name)
    cfg.Framework = "adk"
    cfg.Language = "python"

    // use cfg everywhere below
}
```

## Change Type

/kind cleanup

timflannagan:
@groundnuty Took a quick peek at the code on main. For both of those code blocks you pasted - I don't see `PythonSafeName` defined anywhere in the codebase. At least on main branch. False fire?

timflannagan:
Ah is this related to the https://github.com/agentregistry-dev/agentregistry/pull/205 changes? I see that method being introduced with those changes. Going to close this out in the meantime. Let me know if I'm missing any context here.

groundnuty:
@timflannagan yea sorry, got carried when implementing #205,  can add it to #205 if needed.

PR #205, opened 2026-02-21
fix(cli): Allow hyphens in agent names

### LINKED ISSUE #158: CLI: Allow hypens in agent name with `arctl agent init`
```
$ arctl agent init adk python test-agent
Error: invalid agent name: "test-agent" is not a valid Python identifier: must start with a letter or underscore and contain only letters, digits, and underscores
```

groundnuty:
# Description

Agent names with hyphens (e.g., `my-agent`) are rejected by the Go validator
even though the database constraint already allows them. This is a common
naming convention that users expect to work.

**Root cause:** `agentNameRegex` was `^[a-zA-Z][a-zA-Z0-9]+$`, which only
allowed letters and digits after the initial letter.

**Changes:**
- Relax `agentNameRegex` to `^[a-zA-Z][a-zA-Z0-9-]*[a-zA-Z0-9]$`
- Dots intentionally remain excluded to avoid `.well-known` URL path ambiguity
- Add `PythonSafeName()` helper that converts hyphens to underscores
- Python generator converts agent name before code generation so packages
  remain importable (e.g., `my-agent` -> `my_agent/` directory)

Fixes #158

# Change Type

/kind fix

# Changelog

```release-note
Allow hyphens in agent names, with automatic conversion to underscores for Python code generation
```

# Additional Notes

- Database constraint already allows hyphens - no migration needed
- Dots kept out of validator intentionally (different from DB constraint)
  to avoid routing conflicts with `.well-known` paths
- Only the Python generator converts names; future non-Python frameworks
  can use hyphens directly
- Filed #204 for a pre-existing mutation pattern in `PythonGenerator.Generate`

### DIFF
--- a/internal/cli/agent/frameworks/adk/python/generator.go
+++ b/internal/cli/agent/frameworks/adk/python/generator.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/agentregistry-dev/agentregistry/internal/cli/agent/frameworks/common"
 	"github.com/agentregistry-dev/agentregistry/pkg/models"
+	"github.com/agentregistry-dev/agentregistry/pkg/validators"
 )
 
 //go:embed templates/* templates/agent/* templates/mcp_server/* dice-agent-instruction.md
@@ -30,6 +31,9 @@ func (g *PythonGenerator) Generate(agentConfig *common.AgentConfig) error {
 	if agentConfig == nil {
 		return fmt.Errorf("agent config is required")
 	}
+
+	// Convert agent name to Python-safe identifier (hyphens to underscores)
+	agentConfig.Name = validators.PythonSafeName(agentConfig.Name)
 
 	projectPackageDir := filepath.Join(agentConfig.Directory, agentConfig.Name)
 	if err := os.MkdirAll(projectPackageDir, 0o755); err != nil {
--- a/internal/cli/agent/frameworks/adk/python/generator.go
+++ b/internal/cli/agent/frameworks/adk/python/generator.go
@@ -32,7 +32,9 @@ func (g *PythonGenerator) Generate(agentConfig *common.AgentConfig) error {
 		return fmt.Errorf("agent config is required")
 	}
 
-	// Convert agent name to Python-safe identifier (hyphens to underscores)
+	// Python identifiers cannot contain hyphens (e.g., "my-agent" is parsed as
+	// "my minus agent"), so convert to underscores for the package directory and
+	// module name.
 	agentConfig.Name = validators.PythonSafeName(agentConfig.Name)
 
 	projectPackageDir := filepath.Join(agentConfig.Directory, agentConfig.Name)
--- a/pkg/validators/names.go
+++ b/pkg/validators/names.go
@@ -57,20 +57,22 @@ func ValidateProjectName(name string) error {
 	return nil
 }
 
-// agentNameRegex enforces the strictest rule - names that work BOTH as Python identifiers AND as publishable agent names.
-// Must start with a letter, followed by alphanumeric only, minimum 2 characters.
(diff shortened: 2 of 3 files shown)


shashankram:
Hey @groundnuty, thanks for your contribution. Has this been tested on AWS and GCP?

groundnuty:
@shashankram not really. AWS and GCP aren't mentione in any testing, contributing guidelines in this repo. However I do have access to AWS so I can test, just I need some instructions how/what to test :-)

shashankram:
> @shashankram not really. AWS and GCP aren't mentione in any testing, contributing guidelines in this repo. However I do have access to AWS so I can test, just I need some instructions how/what to test :-)

Sorry my bad. Were you able to test this locally?

internal/cli/agent/frameworks/adk/python/generator.go
```diff
@@ -31,6 +32,9 @@ func (g *PythonGenerator) Generate(agentConfig *common.AgentConfig) error {
 		return fmt.Errorf("agent config is required")
 	}
 
+	// Convert agent name to Python-safe identifier (hyphens to underscores)
```
shashankram: could you update this comment to describe why this is required

pkg/validators/names.go
```diff
@@ -57,20 +57,22 @@ func ValidateProjectName(name string) error {
...
+// Dots are intentionally excluded to avoid .well-known URL path routing ambiguity.
+// Minimum 2 characters.
+var agentNameRegex = regexp.MustCompile(`^[a-zA-Z][a-zA-Z0-9-]*[a-zA-Z0-9]$`)
 
 // ValidateAgentName checks if the agent name is valid.
-// Allowed: letters and digits only, must start with a letter, minimum 2 characters.
-// Not allowed: underscores, dots, hyphens, or Python keywords.
+// Allowed: letters, digits, and hyphens; must start with a letter, end with a letter or digit.
+// Not allowed: underscores, dots, or Python keywords.
 func ValidateAgentName(name string) error {
```
shashankram, line 69: I am confused, because using underscores didn't cause problems. `arctl agent init adk python test-agent` failed but `arctl agent init adk python test_agent` worked. So I am unsure how the latter worked when the comment in L68 suggests otherwise

groundnuty: I tried to validate it, but the CLI build against the PR, and the commit 1ab39e4, the PR is based on, fails when supplying `test_agent` as a name. 
Also in `names-test.go`, there is a specific assessment that tests that `_` is not allowed. 
https://github.com/agentregistry-dev/agentregistry/blob/1ab39e4ddafa81e64afddf50b5874d825d57ff41/pkg/validators/names_test.go#L58
I just did not change it.
Also, `agentNameRegex` in the 1ab39e4 and PR does not allow `_`. 
So I'm also not sure how `_` ever worked.

The comment here restates what tests and regexp show.

internal/cli/agent/frameworks/adk/python/generator.go
groundnuty: I've updated the PR to say
```
  // Python identifiers cannot contain hyphens (e.g., "my-agent" is parsed as
  // "my minus agent"), so convert to underscores for the package directory and
  // module name.
  ```

groundnuty:
@shashankram 
Yes, I've tested it locally:

 -  `make test-unit` — all passing
  - `make lint` — no issues
  - I've built CLI from both base (1ab39e4) and PR (00732c8) commits, verified:
    - `arctl agent init adk python test-agent` — fails on base, succeeds on PR
    - `arctl agent init adk python test_agent` — rejected on both versions 
    - also generated project uses `test_agent/` directory with correct Python package structure

groundnuty:
@shashankram shall we merge it? the bots are threatening to close it :-)

shashankram:
> @shashankram shall we merge it? the bots are threatening to close it :-)

@peterj what do you think?

PR #215, opened 2026-02-25
internal: Materialize skills during agent run

timflannagan:
# Description

Follow-up to #198 that added support for adding metadata to the agent.yaml. This commit adds runtime skill resolution and materialization for `agent run` workflows. When skills have been configured for an agent, we resolve those [ image, registryURL ] input sources to images, and then leverage docker create/cp/etc. to fetch those contents and extract directly to disk.

We leverage existing kagent-adk functionality here where it supports reading skills from an envvar so I only had to extend the docker compose template to mount that directory correctly.

# Change Type

/kind feature

# Changelog

```release-note
NONE
```

### DIFF
--- a/internal/cli/agent/run.go
+++ b/internal/cli/agent/run.go
@@ -70,11 +70,10 @@ func runRun(cmd *cobra.Command, args []string) error {
 	return runFromManifest(cmd.Context(), &manifest, version, nil)
 }
 
-// Note: The below implementation may be redundant in most cases.
-// It allows for registry-type MCP server resolution at run-time, but in doing so, it
-// regenerates folders for servers which were already accounted for (i.e. command-type get
-// generated during their `add-cmd` command).
-// This is not a major issue or breaking, but something we could improve in the future.
+// runFromDirectory runs an agent from a local project directory. It resolves
+// registry-type MCP servers at runtime, regenerating folders for servers that
+// may have already been created during their initial add-cmd invocation. This
+// redundancy is acceptable but could be optimized in the future.
 func runFromDirectory(ctx context.Context, projectDir string) error {
 	manifest, err := project.LoadManifest(projectDir)
 	if err != nil {
@@ -86,7 +85,6 @@ func runFromDirectory(ctx context.Context, projectDir string) error {
 		return fmt.Errorf("failed to resolve skills from agent manifest: %w", err)
 	}
 	if err := materializeSkillsForRuntime(
-		ctx,
 		resolvedSkills,
 		skillsDirForAgentConfig(projectDir, manifest.Name, ""),
 		verbose,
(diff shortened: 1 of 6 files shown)


internal/cli/agent/run.go
```diff
@@ -71,86 +71,45 @@ func runRun(cmd *cobra.Command, args []string) error {
 }
 
 // Note: The below implementation may be redundant in most cases.
-// It allows for registry-type MCP server resolution at run-time, but in doing so, it regenerates folders for servers which were already accounted for (i.e. command-type get generated during their `add-cmd` command)
+// It allows for registry-type MCP server resolution at run-time, but in doing so, it
+// regenerates folders for servers which were already accounted for (i.e. command-type get
+// generated during their `add-cmd` command).
```
timflannagan: TODO: I think this comment is potentially stale?

internal/cli/agent/run.go
```diff
@@ -71,86 +71,45 @@ func runRun(cmd *cobra.Command, args []string) error {
...
+	); err != nil {
+		return fmt.Errorf("failed to materialize skills: %w", err)
+	}
+
 	// Always clear previously resolved registry artifacts to avoid stale folders.
 	if err := project.CleanupRegistryDir(projectDir, verbose); err != nil {
 		return fmt.Errorf("failed to clean registry directory: %w", err)
 	}
 
-	var serversForConfig []common.PythonMCPServer
```
timflannagan, line 87: No changes to this code, just moved to it's own function to be consistent with the new skill runtime code and make this function a bit easier to read.

internal/cli/agent/run.go
```diff
@@ -188,6 +147,86 @@ func hasRegistryServers(manifest *models.AgentManifest) bool {
 	return false
 }
 
+func resolveMCPServersForRuntime(manifest *models.AgentManifest) ([]models.McpServerType, []common.PythonMCPServer, error) {
```
timflannagan: TODO: Move this code to it's own file, e.g. mcp_runtime.go, so we can add direct unit tests? Would be consistent with the new skill runtime code being added here.

internal/cli/agent/run_test.go
```diff
@@ -98,6 +98,61 @@ func TestHasRegistryServers(t *testing.T) {
 	}
 }
 
+func TestResolveMCPServersForRuntime(t *testing.T) {
```
timflannagan, line 101: Added a light unit test for the new extracted function while we're here.

internal/cli/agent/skills_runtime.go
```diff
@@ -0,0 +1,401 @@
...
+		}
+		return apiClient.GetSkillByNameAndVersion(skillName, version)
+	}
+
+	baseURL, err := normalizeSkillRegistryURL(registryURL)
+	if err != nil {
+		return nil, err
+	}
+
+	client := arclient.NewClient(baseURL, "")
```
timflannagan, line 98: Having to define a client this far down into the stack flow seems wrong vs. DI with parameters.

internal/cli/agent/skills_runtime.go
```diff
@@ -0,0 +1,401 @@
...
+	out = replacer.Replace(out)
+	for strings.Contains(out, "--") {
+		out = strings.ReplaceAll(out, "--", "-")
+	}
+	out = strings.Trim(out, "-")
+	if out == "" {
+		return "skill"
+	}
+	return out
+}
```
timflannagan, line 195: AI generated code

AGENTS.md
```diff
@@ -381,6 +381,7 @@ When working with this codebase, AI assistants should:
 4. **Use interfaces** - Define interfaces for new dependencies
 5. **Keep changes minimal** - Only modify what's necessary for the task
 6. **Check database access** - Ensure database operations go through proper layers
+7. **Target upstream for PRs** - When opening pull requests, default to the canonical upstream repository unless the user explicitly asks for a fork target
```
timflannagan: Whoops - I'll remove this and add to my global instructions instead.

internal/cli/agent/skills_runtime.go
```diff
@@ -0,0 +1,401 @@
...
+	if skillResp == nil {
+		return "", fmt.Errorf("skill response is required")
+	}
+	for _, pkg := range skillResp.Skill.Packages {
+		typ := strings.ToLower(strings.TrimSpace(pkg.RegistryType))
+		if (typ == "docker" || typ == "oci") && strings.TrimSpace(pkg.Identifier) != "" {
+			return strings.TrimSpace(pkg.Identifier), nil
+		}
+	}
+	return "", fmt.Errorf("no docker/oci package found")
```
peterj, line 139: kagent will be able to support skills from github repos too (https://github.com/kagent-dev/kagent/pull/1365); so we'll have to update this. 

perhaps add a // TODO here to update this -- I think we'll be then able to check both oci and/or github repo from the registry entry and pick one that's available (or default to one if both are provided).

internal/cli/agent/skills_runtime.go
```diff
@@ -0,0 +1,401 @@
...
+	}
+
+	resp, err := client.GetSkillByNameAndVersion(skillName, version)
+	if err != nil {
+		return nil, fmt.Errorf("fetch skill %q version %q from %s: %w", skillName, version, baseURL, err)
+	}
+	return resp, nil
+}
+
+func normalizeSkillRegistryURL(registryURL string) (string, error) {
```
peterj, line 114: I think we might have this func somewhere already

internal/cli/agent/skills_runtime.go
```diff
@@ -0,0 +1,401 @@
...
+	}
+
+	containerID := strings.TrimSpace(string(output))
+	if containerID == "" {
+		return "", fmt.Errorf("docker create returned empty container id")
+	}
+	return containerID, nil
+}
+
+func runDockerCommand(ctx context.Context, verbose bool, args ...string) error {
```
peterj: we have docker helperes here: internal/cli/common/docker/docker.go -- perhaps use that and/or add stuff to that one if anything is missing.

internal/cli/agent/skills_runtime.go
```diff
@@ -0,0 +1,401 @@
...
+			return false, nil
+		}
+		return false, fmt.Errorf("docker image inspect failed: %w", err)
+	}
+	return true, nil
+}
+
+func copyNonEmptyContents(src, dst string) error {
+	// Skip system directories created by Docker extraction from non-scratch images.
+	skipDirs := map[string]struct{}{
```
peterj: remove from https://github.com/agentregistry-dev/agentregistry/blob/1952e0dadd53a4490a431522834a176573a1a121/internal/cli/skill/pull.go#L136 -- looks like it's the same

internal/cli/agent/skills_runtime.go
```diff
@@ -0,0 +1,401 @@
...
+			continue
+		}
+		if err := copyFile(srcPath, dstPath); err != nil {
+			return fmt.Errorf("copy file %s: %w", srcPath, err)
+		}
+	}
+	return nil
+}
+
+func hasNonEmptyContent(dir string, skipDirs map[string]struct{}) bool {
```
peterj: this exists in pull.go

internal/cli/agent/skills_runtime.go
timflannagan: Any idea which function? I see normalizeBaseURL defined but the logic is slightly different.

internal/cli/agent/skills_runtime.go
timflannagan: Yeah good call. I'll refactor to use that package

PR #219, opened 2026-02-25
fix: three bug fixes for MCP tools and background embedding generation

### LINKED ISSUE #217: Bug: Three issues in MCP tool handlers and background embedding generation
## Summary

Found three bugs while integrating AgentRegistry into a production platform:

### 1. Background embedding goroutines use bare `context.Background()` — missing auth

**File:** `internal/registry/service/registry_service.go` (lines ~198, ~597)

When generating embeddings asynchronously after server/agent creation, the goroutine creates a `context.Background()` with no auth session. Downstream database operations that check `IsSystemSession()` fail silently, causing embeddings to never be persisted.

**Fix:** Wrap with `auth.WithSystemContext(context.Background())`, matching the existing pattern in `embeddings_sse.go`, `embeddings.go`, and `registry_app.go`.

### 2. Nil `Config` map in deployment MCP tool responses

**File:** `internal/mcp/registryserver/server.go`

When a deployment has no config, `Config` is `nil`. JSON-serializing a nil Go map produces `null`, but MCP tool consumers (LLMs/agents) expect an empty object `{}`. This can also cause nil-pointer panics in downstream consumers that iterate or index the map.

Additionally, `list_deployments` uses case-sensitive comparison for the `resource_type` filter, so `"Server"` vs `"server"` returns no results.

**Fix:** Initialize nil Config maps to `map[string]string{}` before returning. Use `strings.EqualFold` for resource type comparison.

### 3. Missing default port for stdio transport in kagent translator

**File:** `internal/runtime/translation/kagent/kagent_translator.go`

When translating a stdio-based MCP server to a kagent `MCPServer` CR, the deployment port is left at `0` if not explicitly set. This is invalid — the KMCP transport adapter needs a port to listen on. The KMCP CLI defaults to 3000.

**Fix:** Default `spec.Deployment.Port` to `3000` for stdio transport, matching the KMCP CLI default.

## Reproduction

1. Enable `EMBEDDINGS_ENABLED=true` and publish a server — embedding column stays NULL (bug 1)
2. Deploy a server with no config, call `get_deployment` via MCP — Config is `null` (bug 2)
3. Register a stdio MCP server, deploy via kagent translator — port is 0 (bug 3)

syn-zhu:
```
/kind fix
```

# Description

Three bug fixes found while integrating AgentRegistry into a production Kubernetes platform.

1. **`auth.WithSystemContext` for background embedding goroutines** — bare `context.Background()` lacks the system auth session, causing embedding updates to fail silently. Matches the pattern already used in `embeddings_sse.go` and `embeddings.go`.

2. **Nil Config map initialization** — `get_deployment` and `list_deployments` MCP tools return `null` for Config when the map is nil, breaking downstream JSON consumers. Also adds case-insensitive `resource_type` filtering.

3. **Default port 3000 for stdio transport** — the kagent translator leaves `Deployment.Port` at 0 for stdio servers, which is invalid. Defaults to 3000 to match the KMCP CLI.

Fixes #217

## Changes

| File | Change |
|------|--------|
| `internal/registry/service/registry_service.go` | `context.Background()` → `auth.WithSystemContext(context.Background())` in 2 goroutines |
| `internal/mcp/registryserver/server.go` | Init nil Config maps, `strings.EqualFold` for resource type filter |
| `internal/runtime/translation/kagent/kagent_translator.go` | Default port 3000 for stdio transport |

## Test plan

- [ ] Publish a server with `EMBEDDINGS_ENABLED=true` → verify embedding column is populated (not NULL)
- [ ] Call `get_deployment` for a deployment with no config → verify Config is `{}` not `null`
- [ ] Call `list_deployments` with `resource_type=server` (lowercase) → verify results returned
- [ ] Deploy a stdio MCP server via kagent translator → verify port is 3000

```release-note
Fix background embedding generation silently failing due to missing auth context, nil Config map in deployment MCP tool responses, and missing default port for stdio transport in kagent translator.
```

### DIFF
--- a/internal/mcp/registryserver/server.go
+++ b/internal/mcp/registryserver/server.go
@@ -5,6 +5,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	restv0 "github.com/agentregistry-dev/agentregistry/internal/registry/api/handlers/v0"
@@ -369,10 +370,14 @@ func addDeploymentTools(server *mcp.Server, registry service.RegistryService) {
 		}
 		outIdx := 0
 		for _, d := range deployments {
-			if args.ResourceType != "" && d.ResourceType != args.ResourceType {
+			if args.ResourceType != "" && !strings.EqualFold(d.ResourceType, args.ResourceType) {
 				continue
 			}
-			resp.Deployments[outIdx] = *d
+			dep := *d
+			if dep.Config == nil {
+				dep.Config = map[string]string{}
+			}
+			resp.Deployments[outIdx] = dep
 			outIdx++
 		}
 		resp.Deployments = resp.Deployments[:outIdx]
@@ -392,7 +397,11 @@ func addDeploymentTools(server *mcp.Server, registry service.RegistryService) {
 		if err != nil {
 			return nil, models.Deployment{}, err
 		}
-		return nil, *deployment, nil
+		dep := *deployment
+		if dep.Config == nil {
+			dep.Config = map[string]string{}
+		}
+		return nil, dep, nil
 	})
 
 	// Deploy server
--- a/internal/registry/service/registry_service.go
+++ b/internal/registry/service/registry_service.go
@@ -17,6 +17,7 @@ import (
 	"github.com/agentregistry-dev/agentregistry/internal/runtime/translation/kagent"
 	"github.com/agentregistry-dev/agentregistry/internal/runtime/translation/registry"
 	"github.com/agentregistry-dev/agentregistry/pkg/models"
+	"github.com/agentregistry-dev/agentregistry/pkg/registry/auth"
 	"github.com/agentregistry-dev/agentregistry/pkg/registry/database"
 	"github.com/jackc/pgx/v5"
 	apiv0 "github.com/modelcontextprotocol/registry/pkg/api/v0"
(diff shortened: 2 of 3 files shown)


syn-zhu:
Hi, @ilackarms gentle bump on this :)

Should be a super quick review!

internal/mcp/registryserver/server.go
```diff
@@ -369,10 +370,14 @@ func addDeploymentTools(server *mcp.Server, registry service.RegistryService) {
 		}
 		outIdx := 0
 		for _, d := range deployments {
-			if args.ResourceType != "" && d.ResourceType != args.ResourceType {
+			if args.ResourceType != "" && !strings.EqualFold(d.ResourceType, args.ResourceType) {
```
inFocus7, line 373: nit: i'm personally a bit iffy on this. it makes sense, but as far as I could tell this is the only place in the code we'd support case-insensitive

internal/mcp/registryserver/server.go
```diff
@@ -369,10 +370,14 @@ func addDeploymentTools(server *mcp.Server, registry service.RegistryService) {
...
 				continue
 			}
-			resp.Deployments[outIdx] = *d
+			dep := *d
+			if dep.Config == nil {
+				dep.Config = map[string]string{}
+			}
+			resp.Deployments[outIdx] = dep
```
inFocus7, line 380: to avoid a `nil` Config, would it be better to update the list method so it gets instantiated there and we wouldn't need to do this check here?

so in `registry_service.go`, when listing k8s deployments `listKubernetesDeployments`, we'd simply
```go
config := labels
if config == nil {
  config = make(map[string]string)
}
d := {
...
  config: config
...
}
```

for what it's worth, our db listing code already does something similar to this, so we'd essentially be matching its behavior in the k8s listing
```go
		if d.Config == nil {
			d.Config = make(map[string]string)
		}
```

PR #220, opened 2026-02-25
fix: use pure vector similarity for semantic search, wire into MCP tools

### LINKED ISSUE #218: Semantic search with hybrid AND prevents natural-language discovery
## Summary

When `semantic_search=true`, the REST handlers and MCP tools AND the vector similarity results with an `ILIKE` substring filter on the `name` column. This prevents natural-language queries from working for agent/server discovery.

## Problem

The current flow for `GET /v0/servers?search=database+management+tools&semantic_search=true`:

1. Sets `filter.SubstringName = "database management tools"` (ILIKE on name column)
2. Sets `filter.Semantic` with vector similarity
3. Sets `filter.Semantic.HybridSubstring = filter.SubstringName`

The database query AND-s both conditions. Since no server is *named* "database management tools", the ILIKE filter returns zero rows, and the entire query returns nothing — even though vector similarity would have found relevant servers.

This defeats the primary use case for semantic search: discovering tools by describing what you need rather than knowing the exact name.

### The HybridSubstring field suggests this was intended to work differently

The existence of the `HybridSubstring` field on `SemanticSearchOptions` implies the original design intended to support a hybrid mode where both vector similarity and substring matching contribute to results. However, the current implementation AND-s them at the SQL level, which means the substring filter *restricts* rather than *augments* the semantic results.

For hybrid search to be useful, it would need to use OR logic (return results matching *either* vector similarity or substring) or a ranked combination. The current AND logic makes `HybridSubstring` actively harmful — it turns every semantic query into an intersection that returns nothing whenever the search phrase doesn't appear as a substring in the name column. This effectively makes semantic search unusable for its primary use case (natural-language discovery).

## Additional issue: MCP tools don't wire semantic search at all

The MCP tool handlers for `list_agents` and `list_servers` accept `semantic_search` and `semantic_threshold` as input parameters (inherited from the REST input types) but never read them. The semantic fields are dead code in the MCP path, so agents calling these tools can never use semantic search.

## Proposed fix

**REST handlers** (`agents.go`, `servers.go`): Change to `if/else` — when `semantic_search=true`, only use vector similarity. When false, fall back to substring matching.

**MCP tool handlers** (`server.go`): Wire up the `semantic_search` and `semantic_threshold` parameters using the same `if/else` pattern.

This makes semantic and substring search mutually exclusive per request, which matches user intent: you either know the name (substring) or you're describing what you need (semantic).

## Alternative approaches

If hybrid search is desired, an OR-based approach (return results matching EITHER vector similarity OR substring) would preserve both capabilities without breaking natural-language queries. The current `HybridSubstring` mechanism could be reworked to use OR/ranked-union instead of AND. Happy to adjust the implementation if that's preferred.

syn-zhu:
```
/kind fix
```

# Description

Fixes #218 — semantic search currently AND-s vector similarity with an ILIKE substring filter on the name column, which prevents natural-language discovery queries from returning results.

## Problem

A query like `?search=database+management+tools&semantic_search=true` returns nothing because:
1. The ILIKE filter matches against server/agent *names* — no server is named "database management tools"
2. Vector similarity would find relevant results, but the AND with the empty ILIKE result set produces zero rows

This defeats the primary use case for semantic search: discovering tools by describing what you need.

The `HybridSubstring` field on `SemanticSearchOptions` suggests hybrid mode was intended to work, but the AND logic at the SQL level makes it actively harmful — it restricts rather than augments results.

Additionally, the MCP tool handlers for `list_agents` and `list_servers` accept `semantic_search` and `semantic_threshold` parameters but never read them — they are dead code.

### Commit 1: Pure vector similarity for semantic search

Changes REST handlers (`agents.go`, `servers.go`) to use `if/else`:
- `semantic_search=true` → only vector similarity (no SubstringName, no HybridSubstring)
- `semantic_search=false` → substring matching as before

### Commit 2: Wire semantic search into MCP tools

Adds the same `if/else` pattern to MCP tool handlers in `server.go` for both `list_agents` and `list_servers`. Updates tool descriptions to mention semantic search capability.

## Alternative approaches

If hybrid search (vector + substring) is desired, I am happy to:
- Make it an explicit `hybrid_search` mode alongside pure semantic
- Use OR instead of AND so either condition can match
- Add a separate parameter to control the behavior

## Test plan

- [ ] `GET /v0/servers?search=database+management&semantic_search=true` → returns relevant servers by embedding similarity
- [ ] `GET /v0/servers?search=postgres` (no semantic) → returns servers with "postgres" in the name (unchanged behavior)
- [ ] MCP `list_servers` with `semantic_search=true, search="file management"` → returns relevant servers
- [ ] MCP `list_servers` with `search="postgres"` (no semantic) → substring match (unchanged)

```release-note
Fix semantic search to use pure vector similarity instead of AND-ing with substring filter. Wire semantic search parameters into MCP list_servers and list_agents tools.
```

### DIFF
--- a/internal/mcp/registryserver/server.go
+++ b/internal/mcp/registryserver/server.go
@@ -45,7 +45,7 @@ type listAgentsArgs = restv0.ListAgentsInput
 func addAgentTools(server *mcp.Server, registry service.RegistryService) {
 	mcp.AddTool(server, &mcp.Tool{
 		Name:        "list_agents",
-		Description: "List published agents with optional search and pagination",
+		Description: "List published agents with optional search and pagination. Set semantic_search=true for natural-language queries.",
 	}, func(ctx context.Context, _ *mcp.CallToolRequest, args listAgentsArgs) (*mcp.CallToolResult, models.AgentListResponse, error) {
 		filter := &database.AgentFilter{}
 
@@ -56,7 +56,17 @@ func addAgentTools(server *mcp.Server, registry service.RegistryService) {
 			}
 			filter.UpdatedSince = &ts
 		}
-		if args.Search != "" {
+		// When semantic search is active, use pure vector similarity.
+		// Otherwise fall back to substring name matching.
+		if args.Semantic {
+			if args.Search == "" {
+				return nil, models.AgentListResponse{}, fmt.Errorf("semantic_search requires the search parameter")
+			}
+			filter.Semantic = &database.SemanticSearchOptions{
+				RawQuery:  args.Search,
+				Threshold: args.SemanticMatchThreshold,
+			}
+		} else if args.Search != "" {
 			filter.SubstringName = &args.Search
 		}
 		if args.Version != "" {
@@ -118,7 +128,7 @@ type listServersArgs = restv0.ListServersInput
 func addServerTools(server *mcp.Server, registry service.RegistryService) {
 	mcp.AddTool(server, &mcp.Tool{
 		Name:        "list_servers",
-		Description: "List published MCP servers with optional search and pagination",
+		Description: "List published MCP servers with optional search and pagination. Set semantic_search=true for natural-language queries (e.g. 'database management tools').",
 	}, func(ctx context.Context, _ *mcp.CallToolRequest, args listServersArgs) (*mcp.CallToolResult, apiv0.ServerListResponse, error) {
 		filter := &database.ServerFilter{}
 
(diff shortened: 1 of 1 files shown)


syn-zhu:
@ilackarms @peterj can we trigger workflows? Thanks

timflannagan:
Just triggered

timflannagan:
@syn-zhu Thanks for the contribution. Apologies for the delay here. Going to close/re-open to help unstick the e2e job hopefully.

PR #222, opened 2026-02-25
feat: make embedding dimensions configurable via EMBEDDINGS_DIMENSIONS

### LINKED ISSUE #221: EMBEDDINGS_DIMENSIONS env var is locked to 1536, cannot use other embedding providers
## Summary

The `EMBEDDINGS_DIMENSIONS` env var exists in `EmbeddingsConfig` but is hardcoded to only accept `1536` via a validation check in `validate.go`:

```go
if cfg.Embeddings.Dimensions != 1536 {
    return fmt.Errorf("embeddings dimensions must equal 1536 to match database schema (got %d)", cfg.Embeddings.Dimensions)
}
```

This makes it impossible to use embedding providers with different default dimensions, such as:
- **Voyage AI** (`voyage-3`, `voyage-3-lite`): 1024 dimensions
- **Cohere** (`embed-v4.0`): 1024 dimensions
- **OpenAI** (`text-embedding-3-large`): 3072 dimensions

## Root cause

The initial migration (`001_initial_schema.sql`) hardcodes `vector(1536)` for the `semantic_embedding` columns. The validation check exists to prevent a mismatch between the config and the schema, which would cause a pgvector error at insert time.

However, this makes the `EMBEDDINGS_DIMENSIONS` config effectively useless since only one value is allowed.

## Proposed solution

Instead of maintaining separate migration files for each dimension, reconcile the schema dynamically at startup:

1. Remove the hardcoded 1536 validation
2. After running migrations, query `pg_attribute.atttypmod` to get the current vector column dimension
3. If it differs from `EMBEDDINGS_DIMENSIONS`, atomically:
   - Drop HNSW indexes
   - Clear existing embeddings (must be regenerated with new dimension anyway)
   - ALTER both columns to `vector(N)` where N is the configured dimension
   - Recreate HNSW indexes

This is safe because:
- Existing 1536 deployments are unaffected (no-op when dimensions match)
- Clearing embeddings on dimension change is correct (old vectors are incompatible)
- The reconciliation runs once on startup, not per-request
- It uses a single transaction for atomicity

Operators can then switch providers with just env vars:
```
EMBEDDINGS_DIMENSIONS=1024
EMBEDDINGS_PROVIDER=voyageai
EMBEDDINGS_MODEL=voyage-3
```

syn-zhu:
```
/kind feature
```

# Description

Fixes #221 - the `EMBEDDINGS_DIMENSIONS` env var exists but is locked to 1536 by a hardcoded validation check, making it impossible to use embedding providers like Voyage AI (1024) or Cohere (1024).

## Approach

Instead of per-dimension migration files, this adds **startup dimension reconciliation**:

1. **Removes** the hardcoded `!= 1536` validation in `validate.go`
2. **Adds** `ensureVectorDimensions()` that runs after migrations in `NewPostgreSQL`
3. It queries `pg_attribute.atttypmod` to get the current column dimension
4. If config differs from schema, it atomically (single transaction):
   - Drops HNSW indexes
   - Clears existing embeddings (incompatible with new dimension)
   - ALTERs both `servers.semantic_embedding` and `agents.semantic_embedding` to `vector(N)`
   - Recreates HNSW indexes with `vector_cosine_ops`

**Callers pass the configured dimension** (or `0` to skip reconciliation for CLI tools and tests).

## Why this approach

- **Zero migration files** to maintain per provider
- **Any dimension** works, not a whitelist
- **Existing 1536 deployments** are unaffected (no-op when dimensions match)
- **Single env var change** to switch providers:
  ```
  EMBEDDINGS_DIMENSIONS=1024
  EMBEDDINGS_PROVIDER=voyageai
  EMBEDDINGS_MODEL=voyage-3
  ```

## Changes

| File | Change |
|------|--------|
| `config/validate.go` | Remove hardcoded `!= 1536` check |
| `database/postgres.go` | Add `embeddingDimensions` param + `ensureVectorDimensions()` |
| `registry_app.go` | Pass `cfg.Embeddings.Dimensions` |
| `cli/import.go`, `cli/export.go` | Pass `0` (skip reconciliation) |
| `database/testutil.go` | Pass `0` (tests use schema default) |

## Test plan

- [ ] Fresh install with default config (1536) - no reconciliation, schema unchanged
- [ ] Fresh install with `EMBEDDINGS_DIMENSIONS=1024` - schema reconciled to vector(1024)
- [ ] Existing 1536 deployment, change to 1024 - columns altered, embeddings cleared, indexes recreated
- [ ] Existing 1024 deployment, restart with same config - no-op
- [ ] CLI import/export with `EMBEDDINGS_DIMENSIONS=1024` - no reconciliation (passes 0)
- [ ] Embeddings disabled - no reconciliation (dimensions param is 0 when not set)

```release-note
The EMBEDDINGS_DIMENSIONS env var now actually works. Operators can configure any vector dimension to match their embedding provider (e.g., 1024 for Voyage AI, 1536 for OpenAI). The database schema is automatically reconciled on startup when the dimension changes.
```

### DIFF
--- a/internal/registry/api/handlers/v0/servers.go
+++ b/internal/registry/api/handlers/v0/servers.go
@@ -128,8 +128,7 @@ func RegisterServersEndpoints(api huma.API, pathPrefix string, registry service.
 		}, nil
 	})
 
-	var tags []string
-	tags = []string{"servers"}
+	var tags = []string{"servers"}
 
 	// List servers endpoint
 	huma.Register(api, huma.Operation{
--- a/internal/registry/api/handlers/v0/servers_test.go
+++ b/internal/registry/api/handlers/v0/servers_test.go
@@ -22,7 +22,6 @@ import (
 	"github.com/agentregistry-dev/agentregistry/pkg/registry/database"
 	"github.com/danielgtaylor/huma/v2"
 	"github.com/danielgtaylor/huma/v2/adapters/humago"
-	"github.com/jackc/pgx/v5"
 	apiv0 "github.com/modelcontextprotocol/registry/pkg/api/v0"
 	"github.com/modelcontextprotocol/registry/pkg/model"
 	"github.com/stretchr/testify/assert"
@@ -136,8 +135,7 @@ func TestListServersEndpoint(t *testing.T) {
 
 func TestListServersSemanticSearch(t *testing.T) {
 	ctx := context.Background()
-	db := internaldb.NewTestDB(t)
-	ensureVectorExtension(t, db)
+	db := internaldb.NewTestDBWithEmbeddings(t, 3)
 
 	cfg := config.NewConfig()
 	cfg.Embeddings.Enabled = true
@@ -434,15 +432,6 @@ func TestGetServerVersionEndpoint(t *testing.T) {
 			}
 		})
 	}
-}
-
-func ensureVectorExtension(t *testing.T, db database.Database) {
-	t.Helper()
-	err := db.InTransaction(context.Background(), func(ctx context.Context, tx pgx.Tx) error {
-		_, execErr := tx.Exec(ctx, "CREATE EXTENSION IF NOT EXISTS vector")
-		return execErr
-	})
-	require.NoError(t, err, "failed to ensure pgvector extension for tests")
 }
 
 type stubEmbeddingProvider struct {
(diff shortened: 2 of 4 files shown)


syn-zhu:
@ilackarms had to fix a few tests

PR #232, opened 2026-02-27
add e2e cli tests

peterj:
# Description

Adding e2e CLI tests for local and kubernetes targets and fixing a couple of issues - we were setting the Cmd in docker compose to "" when OCI image was used (we shouldn't have been doing that); also, making sure any errors are captured when docker compose fails.

# Change Type
```
/kind feature
/kind fix
```

# Changelog

```release-note
fix: resolve OCI runtime failures for MCP servers deployed without a runtime hint
fix: capture docker compose stderr in error messages for local deployment failures
tests: add CLI e2e tests
```


peterj:
k8s not reachable from within the docker container -- need to fix the compose file somehow to make this work.

scripts/kind/kind-config.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+# this is the default configuration for nodes
+nodes:
+  - role: control-plane
+# Example of port mappings for the control-plane node
+#    extraPortMappings:
+#      - containerPort: 30884
+#        hostPort: 30884
+#      - containerPort: 30885
+#        hostPort: 30885
```
timflannagan, line 43: nit: EOF newline.

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,65 @@
...
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+        with:
+          name: ${{ env.BUILDX_BUILDER_NAME }}
+          use: true
+          driver-opts: network=host
+
+      - name: Install Go tool dependencies
+        run: go install tool
```
timflannagan: I think we can rip this out - tools are embedded with the go command itself.

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,65 @@
+name: E2E Tests
+
+on:
+  pull_request:
+    branches: [main]
+    paths-ignore:
+      - '**/*.md'
```
timflannagan: If we add this workflow as a required job in the branch protection rules, we'll run into issues with markdown only PRs based on the way GH handles the paths-ignore field. IIRC, it'll skip that job and that PR can't be merged automatically.

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,65 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: ${{ env.GO_VERSION }}
+          cache: true
+          cache-dependency-path: go.sum
```
timflannagan, line 25: We can use https://github.com/actions/setup-go/blob/main/action.yml#L7-L8 and point that to the root go.mod instead of having to hardcode the Go version in this workflow. Just makes it easier to bump the Go major version in one place.

timflannagan (approved):
This is a good start - I think some of the tests being added here as still in the unit test realm, but we can iterate here now that we have the foundation in place. I'll hold off on posting review comments for that test bed as a result.

.github/workflows/e2e.yaml
peterj: removed

PR #234, opened 2026-02-27
api nits and refactor deployer adapter

ilackarms:
# Description

- finish rename config -> env
- call deployment adapters from within registry service
- add extraroutes

# Change Type

# Changelog
/kind cleanup
/kind design

```release-note
NONE
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,7 @@ require (
 	github.com/danielgtaylor/huma/v2 v2.34.1
 	github.com/golang-jwt/jwt/v5 v5.3.0
 	github.com/google/go-containerregistry v0.20.6
+	github.com/google/uuid v1.6.0
 	github.com/jackc/pgx/v5 v5.7.6
 	github.com/joho/godotenv v1.5.1
 	github.com/kagent-dev/kagent/go v0.0.0-20251107200645-686008ea62ac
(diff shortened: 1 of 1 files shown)


internal/registry/api/handlers/v0/deployment_adapters.go
```diff
@@ -3,108 +3,128 @@ package v0
...
+	providerID := req.ProviderID
+	if providerID == "" {
+		providerID = "kubernetes-default"
+	}
+	env := req.Env
+	if env == nil {
+		env = map[string]string{}
+	}
+	switch req.ResourceType {
+	case "", "mcp":
```
peterj: I think we should fail here if resource type is not set, what do you think?

internal/registry/api/handlers/v0/deployments.go
```diff
@@ -190,25 +184,29 @@ func RegisterDeploymentsEndpoints(api huma.API, basePath string, registry servic
...
+			Version:          input.Body.Version,
+			ProviderID:       providerID,
+			ResourceType:     resourceType,
+			Status:           "",
+			Origin:           "",
+			Env:              input.Body.Env,
+			ProviderConfig:   input.Body.ProviderConfig,
+			ProviderMetadata: nil,
+			PreferRemote:     input.Body.PreferRemote,
+			DeployedBy:       "",
```
peterj: where would the origin and deployby values come from?

pkg/cli/root.go
```diff
@@ -64,14 +64,6 @@ func Root() *cobra.Command {
 	return rootCmd
 }
 
-// Execute runs the root command with default flags (e.g. verbose) and exits on error.
-func Execute() {
```
peterj, line 68: removed by mistake?

internal/registry/api/handlers/v0/deployment_adapters.go
ilackarms: ah yeah i thought i removed that 👍

internal/registry/api/handlers/v0/deployments.go
ilackarms: deployedbhy supposed to be removed, origin is always managed in kube/local

pkg/cli/root.go
ilackarms: the code was unused afaict

peterj (approved):

PR #240, opened 2026-03-02
api refactor: deployments to ID-based APIs

ilackarms:
# Description

This PR finalizes the deployment API refactor as a clean break, removing compatibility shims and aligning client/server behavior with the new `/v0` design.

- **Motivation:** Eliminate ambiguous deployment identity and deprecated payload shapes, and enforce one canonical contract.
- **What changed:**
  - Deployment create/get/delete flows now consistently use deployment `id` (UUID) as the only identifier.
  - Removed handler-side ID override so IDs come from the service/adapter layer source of truth.
  - Updated client methods to ID-based deployment operations (`GetDeploymentByID`, `RemoveDeploymentByID`), removing legacy name/version deployment paths.
  - Removed deprecated deployment response compatibility fields from client (`config`, `runtime`), keeping canonical `env`.
  - Simplified client create semantics (removed `Push*` compatibility wrappers; `Create*` directly maps to canonical endpoints).
  - Fixed client retry behavior and restored SSE no-timeout behavior for streaming indexing.
  - Added/updated tests to cover ID round-trip behavior and `JSONObject` conversion correctness.

# Change Type

/kind breaking_change  
/kind fix  
/kind cleanup

# Changelog

```release-note
BREAKING: deployment APIs and client operations are now strictly ID-based (`/v0/deployments/{id}`), and deprecated deployment compatibility fields (`config`, `runtime`) have been removed in favor of canonical `env`.
```

# Additional Notes

- This PR intentionally does **not** preserve backwards compatibility for deprecated deployment fields or legacy deployment route shapes.

### DIFF
--- a/internal/registry/api/handlers/v0/embeddings.go
+++ b/internal/registry/api/handlers/v0/embeddings.go
@@ -108,13 +108,15 @@ func registerIndexEndpoint(
 			return nil, huma.Error500InternalServerError("failed to create job: " + err.Error())
 		}
 
+		initialStatus := string(job.Status)
+
 		// Run indexing in background
 		go runIndexJob(indexer, jobManager, job.ID, req)
 
 		return &types.Response[IndexJobResponse]{
 			Body: IndexJobResponse{
 				JobID:  string(job.ID),
-				Status: string(job.Status),
+				Status: initialStatus,
 			},
 		}, nil
 	})


internal/registry/api/handlers/v0/deployments.go
```diff
@@ -185,26 +183,23 @@ func RegisterDeploymentsEndpoints(api huma.API, basePath string, registry servic
 		platform := normalizePlatform(provider.Platform)
 
 		deploymentReq := &models.Deployment{
-			ID:               uuid.NewString(),
```
peterj, line 188: if we're removing the ID here, is there still a unique ID we'll use for deployments?

internal/registry/service/registry_service.go
```diff
@@ -874,7 +911,7 @@ func (s *registryServiceImpl) DeployServer(ctx context.Context, serverName, vers
 		}
 		// Deployment record already exists — clean up stale record and retry
 		log.Printf("Deployment for %s/%s already exists, replacing stale record", serverName, deployment.Version)
-		if cleanupErr := s.cleanupExistingDeployment(ctx, deployment.ID, provider.Platform); cleanupErr != nil {
+		if cleanupErr := s.cleanupExistingDeployment(ctx, serverName, deployment.Version, resourceTypeMCP, provider.Platform); cleanupErr != nil {
```
peterj: if we're not using the ID, does that mean I can only deploy an artifact once -- like one MCPServer or Agent deployment in kubernetes?

internal/registry/api/handlers/v0/deployments.go
ilackarms: yes, deployments still get a unique ID. handler-side ID assignment was removed so the service/storage layer is the source of truth. the db `CreateDeployment` call assigns a UUID when deployment.ID is empty, and that returned ID is what clients use

```go
	if deployment.ID == "" {
		_ = db.getExecutor(tx).QueryRow(ctx, "SELECT uuid_generate_v4()::text").Scan(&deployment.ID)
	}
```

internal/registry/service/registry_service.go
ilackarms: yes although this is already the existing behavior; current db shape still enforces single-row identity from legacy semantics (not ID semantics). supporting parallel deployments across providers would require a schema + service semantics change (we can leave this for the post-api phase of the refactor)

```sql
CREATE TABLE deployments (
    -- Primary identifiers
    server_name VARCHAR(255) NOT NULL,
    version VARCHAR(255) NOT NULL,
    ...
    -- Primary key
    CONSTRAINT deployments_pkey PRIMARY KEY (server_name, version)
);
```

internal/cli/mcp/list.go
```diff
@@ -257,6 +245,36 @@ func printServersTable(servers []*v0.ServerResponse, deployedServers []*client.D
 	}
 }
 
+func buildMCPDeploymentCounts(deployedServers []*client.DeploymentResponse) map[string]map[string]int {
```
peterj: would it make sense to extract these two functions to a common place, since they are more or less the same as the one in agent list?

internal/runtime/translation/kagent/kagent_translator.go
```diff
@@ -318,20 +328,50 @@ func buildRemoteMCPURL(host string, port uint32, path string) string {
...
 	if version != "" {
 		base = fmt.Sprintf("%s-%s", name, version)
 	}
+	if deploymentID != "" {
+		base = fmt.Sprintf("%s-%s", base, deploymentID)
+	}
+	return sanitizeK8sName(base)
+}
+
+func RemoteMCPResourceName(name, deploymentID string) string {
```
peterj, line 335: these two functions are exactly the same

internal/registry/service/registry_service.go
```diff
@@ -810,74 +810,12 @@ func (s *registryServiceImpl) resolveProviderByID(ctx context.Context, providerI
 	return s.db.GetProviderByID(ctx, nil, providerID)
 }
 
-// cleanupKubernetesResources deletes Kubernetes runtime resources for a stale deployment.
-// Errors are logged but not returned, since the resources may already be gone.
-func (s *registryServiceImpl) cleanupKubernetesResources(ctx context.Context, existing *models.Deployment) {
```
peterj, line 790: are we not cleaning these up anymore?

please make sure this scenario works:

1. deploy an agent to k8s (arctl agent deploy)
2. manually delete the agent from k8s
3. deploy the agent again to k8s (arctl agent deploy)

ilackarms: we no longer do name/version cleanup on deploy, because that breaks once multiple deployments can exist for the same registry artifact. deployments are now 1:1 by deployment ID, and k8s resources are named/labeled with that ID.

for your scenario (deploy -> manually delete in k8s -> deploy again), this should work. redeploy creates a new deployment ID and does not collide with previous runtime state.

note that deleting the deployment from the DB if it is deleted directly from k8s will require a follow up PR (as we discussed to implement db reconciliation feature)

i plan to address this once the api refactor is done

peterj (approved):

Issue #227, opened 2026-02-26
Fix "vlatest" output when publishing agents/servers with non-semver versions


timflannagan:
When publishing an agent with the "latest" version, I'm seeing the following output locally:

```bash
$ arctl agent publish /work/testagent
Publishing agent: testagent (vlatest)
✓ Published: testagent (vlatest)
```

That said, this doesn't seem to have any effect on the actual agent versioning:

```bash
$ arctl agent list
NAME        VERSION   FRAMEWORK   LANGUAGE   PROVIDER   MODEL              DEPLOYED
testagent   latest    adk         python     gemini     gemini-2.0-flash   False
```

So just low hanging fruit / UX issue.

timflannagan:
Also relevant for the agent delete command too

```bash
$ arctl agent delete testagent --version latest                                                                          1 ↵
Deleting agent testagent version latest...
✓ Deleted: testagent (vlatest)
```

timflannagan:
Looks like I fixed the duplicate v, but the original issue is still reproducible:

```bash
$ ./bin/arctl agent publish /work/testagent --version latest --overwrite
Publishing agent: testagent (vlatest)
Overwriting existing agent testagent version latest
✓ Published: testagent (vlatest)
```

Re-opening since it should be a quick fix.

PR #247, opened 2026-03-03
internal: Fix duplicate v in CLI version display

timflannagan:
# Description

This fixes CLI output formatting that could render semantic versions with a duplicated prefix (for example, `vv0.0.1`) when the input version already included `v`.

```bash
$ /bin/arctl skill publish --github https://github.com/antfu/skills/tree/main/skills/vue vue --version v0.0.1
✓ Published: vue (vv0.0.1)
```

The change centralizes version display normalization in `internal/cli/common` and updates publish, delete, deploy, and agent list output paths to use it consistently.

Fixes https://github.com/agentregistry-dev/agentregistry/issues/227

# Change Type

/kind fix

# Changelog

```release-note
Fix duplicate `v` prefix in CLI version output across publish, delete, deploy, and list commands.
```

### DIFF
--- a/internal/cli/common/common.go
+++ b/internal/cli/common/common.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"github.com/agentregistry-dev/agentregistry/internal/cli/mcp/manifest"
+	versionpkg "github.com/agentregistry-dev/agentregistry/internal/version"
 	"github.com/agentregistry-dev/agentregistry/pkg/printer"
 	"github.com/stoewer/go-strcase"
 )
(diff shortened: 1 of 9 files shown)


internal/cli/common/common.go
```diff
@@ -76,3 +77,8 @@ func ResolveVersion(flagVersion, manifestVersion string) string {
 	}
 	return "latest"
 }
+
+// FormatVersionForDisplay normalizes version display with a leading "v".
+func FormatVersionForDisplay(version string) string {
+	return versionpkg.EnsureVPrefix(version)
+}
```
timflannagan, line 84: A bit awkward, but thought this approach was better for centralizing a way for other CLI commands to call a function that lives in this package vs. calling the exported function in the version package directly.

PR #271, opened 2026-03-05
fix: auto-generate otel collector config when telemetryEndpoint is set

### LINKED ISSUE #176: agent run fails if you manually add telemetryEndpoint to agent.yaml
I manually added `telemetryEndpoint` to agent.yaml and when doing `arctl agent run`, I get:

```
Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error mounting "/host_mnt/Users/peterj/projects/agentregistry/agentdemo/otel-collector-config.yaml" to rootfs at "/etc/otelcol/config.yaml": create mountpoint for /etc/otelcol/config.yaml mount: cannot create subdirectories in "/var/lib/docker/rootfs/overlayfs/72d73768eda544b94b0c95594a128f75c807d62c081d8f8e1de4ab7e3ff34348/etc/otelcol/config.yaml": not a directory: unknown: Are you trying to mount a directory onto a file (or vice-versa)? Check if the specified host path exists and is the expected type
Error: failed to start docker compose: exit status 1
Usage:
  arctl agent run [project-directory-or-agent-name] [flags]

Examples:
```

It tries to mount in the otel-collector-config.yaml (which doesn't exist). 

We need to make sure before we run that the available files are in the folder based on the config. Since we can't prevent users from manually modifying the agent.yaml we can either show them a better error OR re-provision the files based on the config they've updated.

optimus-fulcria:
# Description

Auto-generate the OpenTelemetry collector config file when a user sets `telemetryEndpoint` in agent.yaml but has not run `arctl agent init --telemetry`.

**What changed:**
- Added `ensureOtelCollectorConfig()` that generates a default collector config from an embedded template when the file is missing
- Applied to both `runFromDirectory` and `runFromManifest` code paths
- 3 table-driven tests added

Fixes #176

# Change Type

/kind fix

# Changelog

```release-note
Auto-generate otel collector config when telemetryEndpoint is set but config file is absent
```

# Additional Notes

Preserves existing custom configs — only writes if the file does not exist.


internal/cli/agent/run.go
```diff
@@ -580,3 +588,36 @@ func buildRegistryResolvedServers(tempDir string, manifest *models.AgentManifest
...
 	return nil
 }
+
+const otelCollectorConfigFile = "otel-collector-config.yaml"
+
+// ensureOtelCollectorConfig generates the OpenTelemetry collector config file
+// when the manifest has a telemetryEndpoint but the file is missing. This
+// handles the case where a user manually adds telemetryEndpoint to agent.yaml
+// without going through arctl agent init --telemetry.
+func ensureOtelCollectorConfig(dir string, manifest *models.AgentManifest) error {
```
peterj: this is good, but I think it would make more sense if it was in the internal/cli/agent/project/project.go as that's where we deal with generating/regenerating project files. Perhaps `EnsureOtelCollectorConfig` or something like that; and same signature as the `Regenerate*` functions in that file.

peterj (changes requested):

peterj (approved):

PR #274, opened 2026-03-05
fix: three bugs in MCP tool handlers and kagent translator

### LINKED ISSUE #217: Bug: Three issues in MCP tool handlers and background embedding generation

optimus-fulcria:
# Description

Fix three bugs in MCP tool handlers and the kagent translator.

**What changed:**
1. Background embedding goroutines now use `auth.WithSystemContext()` instead of bare `context.Background()` (fixes auth failures)
2. Nil Config maps initialized to empty maps; `resource_type` filter is now case-insensitive
3. Stdio transport defaults port to 3000 (KMCP CLI default) when not specified

Fixes #217

# Change Type

/kind fix

# Changelog

```release-note
Fix auth context in background goroutines, case-insensitive resource_type filter, and stdio default port
```

# Additional Notes

Added tests for case-insensitive filter and stdio default port.

### DIFF
--- a/internal/mcp/registryserver/deployments_test.go
+++ b/internal/mcp/registryserver/deployments_test.go
@@ -301,6 +301,63 @@ func TestDeploymentTools_FilterResourceType(t *testing.T) {
 	assert.Equal(t, "com.example/echo-agent", out.Deployments[0].ServerName)
 }
 
+func TestDeploymentTools_FilterResourceTypeCaseInsensitive(t *testing.T) {
+	ctx := context.Background()
+
+	deployments := []*models.Deployment{
+		{
+			ID:           "dep-mcp",
+			ServerName:   "com.example/weather",
+			ResourceType: "mcp",
+			Env:          map[string]string{},
+		},
+		{
+			ID:           "dep-agent",
+			ServerName:   "com.example/echo-agent",
+			ResourceType: "agent",
+			Env:          map[string]string{},
+		},
+	}
+
+	reg := servicetesting.NewFakeRegistry()
+	reg.GetDeploymentsFn = func(ctx context.Context, filter *models.DeploymentFilter) ([]*models.Deployment, error) {
+		return deployments, nil
+	}
+
+	server := NewServer(reg)
+	clientTransport, serverTransport := mcp.NewInMemoryTransports()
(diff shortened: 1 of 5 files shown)


optimus-fulcria:
Closing in favor of #219 by @syn-zhu which addresses the same three bugs. Thanks!

PR #279, opened 2026-03-06
split out skill build command from skill publish

### LINKED ISSUE #243: add `arctl skill build` command
Follow the same pattern as with the MCP servers where we have separate build command with the `--push` flag, so it's not part of the `arctl skill publish`

peterj:
# Description

Extracts out the 'build' functionality from the `skill publish` command into a separate command (just like we have it with agents and MCP servers).

Docs: https://github.com/agentregistry-dev/website/pull/13 

Fixes  #243 

# Change Type

```
/kind feature
```

# Changelog

```release-note
added `arctl skill build` command
```

### DIFF
--- a/internal/cli/skill/publish.go
+++ b/internal/cli/skill/publish.go
@@ -268,22 +268,23 @@ func parseSkillFrontmatter(skillPath string) (*skillFrontmatter, error) {
 		return nil, fmt.Errorf("failed to parse SKILL.md frontmatter: %w", err)
 	}
 
+	if fm.Name == "" {
+		return nil, fmt.Errorf("SKILL.md frontmatter missing required field: name")
+	}
+	if fm.Description == "" {
+		return nil, fmt.Errorf("SKILL.md frontmatter missing required field: description")
+	}
+
 	return &fm, nil
 }
 
-// resolveSkillMeta parses SKILL.md frontmatter and resolves the skill name.
+// resolveSkillMeta parses SKILL.md frontmatter and returns the skill name and description.
 func resolveSkillMeta(skillPath string) (name, description string, err error) {
 	fm, err := parseSkillFrontmatter(skillPath)
 	if err != nil {
 		return "", "", err
 	}
-
-	name = fm.Name
-	if name == "" {
-		name = filepath.Base(skillPath)
-	}
-
-	return name, fm.Description, nil
+	return fm.Name, fm.Description, nil
 }
 
 // resolveGitHubVersion returns the version for a GitHub-based publish.
--- a/internal/cli/skill/publish_test.go
+++ b/internal/cli/skill/publish_test.go
@@ -34,24 +34,24 @@ Some content here.
 			wantDesc: "A test skill",
 		},
 		{
-			name: "name only",
+			name: "name only (missing description)",
 			content: `---
 name: simple-skill
 ---
 Body text.
 `,
-			wantName: "simple-skill",
-			wantDesc: "",
+			wantErr:     true,
+			errContains: "missing required field: description",
 		},
 		{
-			name: "empty name falls through",
+			name: "description only (missing name)",
 			content: `---
 description: no name provided
 ---
 Body.
 `,
-			wantName: "",
-			wantDesc: "no name provided",
+			wantErr:     true,
+			errContains: "missing required field: name",
 		},
 		{
 			name:        "empty file",
@@ -160,17 +160,6 @@ description: A skill
 			wantRepoURL: "https://github.com/org/repo",
 		},
 		{
-			name: "falls back to directory name when name is empty",
-			skillMd: `---
-description: No name
----
-`,
-			github:      "https://github.com/org/repo",
-			version:     "2.0.0",
-			wantVer:     "2.0.0",
-			wantRepoURL: "https://github.com/org/repo",
-		},
-		{
 			name: "full tree URL with branch and path",
 			skillMd: `---
 name: nested-skill
@@ -213,11 +202,8 @@ description: Branch
 				t.Fatalf("unexpected error: %v", err)
 			}
 
-			if tt.wantName != "" && skill.Name != tt.wantName {
+			if skill.Name != tt.wantName {
 				t.Errorf("Name = %q, want %q", skill.Name, tt.wantName)
-			}
-			if tt.wantName == "" && skill.Name == "" {
-				t.Error("expected Name to fall back to directory name, got empty")
 			}
 			if skill.Version != tt.wantVer {
 				t.Errorf("Version = %q, want %q", skill.Version, tt.wantVer)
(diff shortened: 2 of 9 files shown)


internal/cli/skill/build.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	cmd := exec.Command("docker", args...)
+	cmd.Dir = skillPath
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	cmd.Stdin = strings.NewReader("FROM scratch\nCOPY . .\n")
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("docker build failed for skill %q: %w", name, err)
+	}
```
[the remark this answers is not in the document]
peterj: fixing.

internal/cli/skill/publish.go
```diff
@@ -91,39 +89,23 @@ func runPublish(cmd *cobra.Command, args []string) error {
...
-	isFolder := false
 	if info, err := os.Stat(absPath); err == nil && info.IsDir() {
-		if _, detectErr := detectSkills(absPath); detectErr == nil {
-			isFolder = true
+		skills, detectErr := detectSkills(absPath)
+		if detectErr != nil {
+			if verbose {
+				printer.PrintInfo(fmt.Sprintf("Directory %q is not a valid skill folder: no SKILL.md found", absPath))
+			}
+			return fmt.Errorf("directory %q does not contain a SKILL.md file (or any immediate subdirectory with one)", input)
```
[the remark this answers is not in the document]
peterj: fixing

internal/cli/skill/build.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+var (
+	buildImage    string
+	buildPush     bool
+	buildPlatform string
+)
+
+func init() {
+	BuildCmd.Flags().StringVar(&buildImage, "image", "", "Full image specification (e.g., docker.io/myorg/my-skill:v1.0.0)")
```
inFocus7, line 42: nit: we could do a `BuildCmd.MarkFlagRequired("image")`

internal/cli/skill/build.go
```diff
@@ -0,0 +1,125 @@
...
+	BuildCmd.Flags().BoolVar(&buildPush, "push", false, "Push the image to the registry")
```
inFocus7: Could we update this description to something like "Push the image to the container registry specified by --image"?

It's technically true, but I personally got confused for a second thinking this referred to _agentregistry_ registry, not the registry the image is created for.

I think we use this wording in our other `--push` flags, so this request isn't blocking, but it could improve understandability.

internal/cli/skill/build.go
```diff
@@ -0,0 +1,125 @@
...
+	dockerExec := docker.NewExecutor(verbose, "")
+	if err := dockerExec.CheckAvailability(); err != nil {
+		return fmt.Errorf("docker check failed: %w", err)
+	}
+
+	for _, skillPath := range skills {
+		if err := buildSkillImage(skillPath, dockerExec); err != nil {
+			return err
+		}
+	}
```
inFocus7: q: since we only take in a single `image` flag and using it, are we just overriding the same image, essentially making it so only the last found skill is built/pushed?

internal/cli/skill/publish.go
```diff
@@ -40,41 +36,43 @@ var PublishCmd = &cobra.Command{
...
+3. Direct registration with a pre-built Docker image:
+   arctl skill publish my-skill \
+     --docker-image docker.io/myorg/my-skill:v1.0.0 \
+     --version 1.0.0 \
+     --description "My Docker skill"
+
+For GitHub modes, SKILL.md must exist at the specified GitHub path.
 In folder mode, the local skill folder must also contain a SKILL.md file with proper YAML frontmatter.
-If the path contains multiple subdirectories with SKILL.md files, all will be published.`,
+If the given directory itself does not contain a SKILL.md but has immediate subdirectories
```
inFocus7: maybe i don't understand the use-case or underlying arch, although what's the purpose of multi-skill publishing?

since we require a `docker-image` (or `github`, but haven't tried this path) which is a singular reference, then aren't we publishing multiple skills using the same docker image?

internal/cli/skill/build.go
peterj: good point! we need to pick the first one only.

internal/cli/skill/publish.go
peterj: needs to be just one -- fixing this.

internal/cli/skill/publish_test.go
```diff
@@ -1265,6 +1148,104 @@ func TestCheckGitHubSkillMdExists_VerifiesCorrectPath(t *testing.T) {
...
+			},
+			want: true,
+		},
+		{
+			name: "valid SKILL.md with description only (no name)",
+			setup: func(dir string) {
+				writeFile(t, filepath.Join(dir, "SKILL.md"), "---\ndescription: no name\n---\nBody.\n")
+			},
+			want: true,
+		},
```
inFocus7, line 1165: nit: we should probably make `name` and `description` a requirement per the agent skill spec ([1](https://agentskills.io/specification#frontmatter-required), [2](https://platform.claude.com/docs/en/agents-and-tools/agent-skills/overview#skill-structure)).

unless we don't want to be in the field(?) of skill validation, then understandable not to require them

peterj: fixed

internal/cli/skill/build.go
```diff
@@ -0,0 +1,122 @@
...
+
+var (
+	buildImage    string
+	buildPush     bool
+	buildPlatform string
+)
+
+func init() {
+	BuildCmd.Flags().StringVar(&buildImage, "image", "", "Full image specification (e.g., docker.io/myorg/my-skill:v1.0.0)")
+	BuildCmd.Flags().BoolVar(&buildPush, "push", false, "Push the image to the container registry, specififed by --image")
```
timflannagan, line 43: specififed typo, but that seems to also be applicable to the other build commands too.

internal/cli/skill/build.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	dockerExec := docker.NewExecutor(verbose, "")
+	if err := dockerExec.CheckAvailability(); err != nil {
+		return fmt.Errorf("docker check failed: %w", err)
+	}
+
+	if err := buildSkillImage(absPath, dockerExec); err != nil {
+		return err
+	}
```
timflannagan, line 73: I missed this during the first review. We initialize a docker executor twice. One here and again on line 109. We already DI the docker executor in buildSkillImage.

internal/cli/skill/build.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+var (
+	buildImage    string
+	buildPush     bool
+	buildPlatform string
+)
+
+func init() {
+	BuildCmd.Flags().StringVar(&buildImage, "image", "", "Full image specification (e.g., docker.io/myorg/my-skill:v1.0.0)")
```
timflannagan, line 42: It looks like we have inconsistent image shorthands across this command, MCP, and agent. MCP uses -n.

PR #287, opened 2026-03-06
fix: include resource name in deployment not-found error

optimus-fulcria:
# Description

When a deployment is not found, the error message says "Resource not found in registry" without specifying which resource. This change uses `err.Error()` for `ErrNotFound` so the service layer's contextual message (which includes the resource name) is surfaced to the user.

**Before:**
```
Error: Resource not found in registry
```

**After:**
```
Error: MCP server 'my-server' not found in registry
```

# Change Type

/kind fix

# Changelog

```release-note
Include resource name in deployment not-found error messages
```

# Additional Notes

- Conflicts with upstream auth error handling resolved: kept separate `ErrUnauthenticated`/`ErrForbidden` cases from upstream, only changed `ErrNotFound` to use `err.Error()`
- Same pattern applied to `removeDeploymentHTTPError`

### DIFF
--- a/internal/registry/api/handlers/v0/deployments.go
+++ b/internal/registry/api/handlers/v0/deployments.go
@@ -83,7 +83,7 @@ func createDeploymentHTTPError(err error) error {
 	case errors.Is(err, database.ErrInvalidInput):
 		return huma.Error400BadRequest(err.Error())
 	case errors.Is(err, database.ErrNotFound):
-		return huma.Error404NotFound("Resource not found in registry")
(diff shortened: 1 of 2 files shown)


peterj (changes requested):
please resolve conflicts and  fill out the PR template

optimus-fulcria:
Conflicts resolved and PR template updated. Kept the upstream separate auth error handling while surfacing the resource name in ErrNotFound via `err.Error()`.

PR #288, opened 2026-03-06
fix: unlock EMBEDDINGS_DIMENSIONS from hardcoded 1536

### LINKED ISSUE #221: EMBEDDINGS_DIMENSIONS env var is locked to 1536, cannot use other embedding providers

optimus-fulcria:
# Description

**Motivation:** The `EMBEDDINGS_DIMENSIONS` env var exists but is locked to `1536` by a validation check, making it impossible to use embedding providers with different default dimensions (Voyage AI 1024, Cohere 1024, OpenAI text-embedding-3-large 3072).

**What changed:**
- Removed the hardcoded `== 1536` validation in `validate.go`, replaced with a reasonable upper bound (16384)
- Added `ReconcileEmbeddingDimensions` to the database layer that runs at startup when embeddings are enabled
- On dimension mismatch: drops HNSW indexes, clears stale embeddings, alters vector columns, recreates indexes — all in a single transaction
- When dimensions match (default 1536 case): no-op

Fixes #221

# Change Type

/kind fix

# Changelog

```release-note
EMBEDDINGS_DIMENSIONS is no longer locked to 1536. Operators can now configure any dimension to match their embedding provider (e.g., 1024 for Voyage AI, 3072 for OpenAI text-embedding-3-large). The database schema is reconciled automatically at startup.
```

# Additional Notes

- Existing 1536 deployments are completely unaffected (reconciliation is a no-op when dimensions match)
- Clearing embeddings on dimension change is correct — old vectors are incompatible with new dimensions and must be regenerated
- Added unit tests for validation and integration tests for the reconciliation logic

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -58,6 +58,9 @@ AGENT_REGISTRY_AGENT_GATEWAY_PORT=8081
 AGENT_REGISTRY_EMBEDDINGS_ENABLED=false
 AGENT_REGISTRY_EMBEDDINGS_PROVIDER=openai
 AGENT_REGISTRY_EMBEDDINGS_MODEL=text-embedding-3-small
+# Dimensions must match your embedding provider/model (e.g., 1536 for OpenAI text-embedding-3-small,
+# 1024 for Voyage AI voyage-3, 3072 for OpenAI text-embedding-3-large).
+# On startup the server reconciles the database schema to match this value.
 AGENT_REGISTRY_EMBEDDINGS_DIMENSIONS=1536
 AGENT_REGISTRY_EMBEDDINGS_ON_PUBLISH=false
 AGENT_REGISTRY_OPENAI_API_KEY=
--- a/internal/registry/config/validate.go
+++ b/internal/registry/config/validate.go
@@ -12,9 +12,8 @@ func Validate(cfg *Config) error {
 		if cfg.Embeddings.Dimensions <= 0 {
 			return fmt.Errorf("embeddings dimensions must be positive (got %d)", cfg.Embeddings.Dimensions)
 		}
-		// Database schema currently provisions vector(1536). Reject mismatches early.
-		if cfg.Embeddings.Dimensions != 1536 {
-			return fmt.Errorf("embeddings dimensions must equal 1536 to match database schema (got %d)", cfg.Embeddings.Dimensions)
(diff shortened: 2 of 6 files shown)


optimus-fulcria:
Closing in favor of #222 by @syn-zhu which addresses the same configurable embedding dimensions feature. Thanks!

PR #290, opened 2026-03-06
fix: pre-populate registry URL in add-skill, add-mcp, and TUI wizard

### LINKED ISSUE #145: pre-populate the registry URL where it makes sense ( in TUI, the arctl agent [add-mcp/add-skill/add-prompt] commands)
For the commands that take in the registry URL value, we should make those flags optional and then just use whatever the default registry URL is (checking the environment variable).

This also applies for the `agent add-mcp` command, in the TUI where we ask the user to enter the URL.

optimus-fulcria:
# Description

Fixes #145

When using `arctl agent add-mcp` or `arctl agent add-skill` with a registry MCP server, the user has to manually type the registry URL even though there's already a configured default. This pre-populates the registry URL field in both the TUI wizard and CLI flags.

## Key changes
- TUI wizard's `chooseRegistryType()` pre-populates the registry URL input with `GetDefaultRegistryURL()`
- CLI `add-mcp` falls back to default registry URL when `--registry-url` is not provided but `--registry-server-name` is
- New `agentutils.GetDefaultRegistryURL()` helper reads from config

# Change Type

/kind fix

# Changelog

```release-note
Pre-populate registry URL in add-skill, add-mcp, and TUI wizard from configured default
```

# Additional Notes

None.

### DIFF
--- a/internal/cli/agent/add-mcp.go
+++ b/internal/cli/agent/add-mcp.go
@@ -74,7 +74,7 @@ func addMcpCmd(name string) error {
 
 	// If flags provided, build non-interactively; else run wizard
 	var res models.McpServerType
-	if remoteURL != "" || command != "" || image != "" || build != "" || registryURL != "" || registryServerName != "" {
+	if remoteURL != "" || command != "" || image != "" || build != "" || registryURL != "" || registryServerName != "" { //nolint:nestif
 		if remoteURL != "" {
 			headerMap, err := utils.ParseKeyValuePairs(headers)
 			if err != nil {
--- a/internal/cli/agent/tui/mcp_wizard.go
+++ b/internal/cli/agent/tui/mcp_wizard.go
@@ -866,11 +866,7 @@ func (w *McpServerWizard) chooseRegistryType() {
 	w.chosenType = serverTypes.Registry.ID
 	w.step = stepRegistryURL
 	// Pre-populate with the current registry URL if available
-	if defaultURL := agentutils.GetDefaultRegistryURL(); defaultURL != "" {
-		w.registryURLInput.SetValue(defaultURL)
-	} else {
-		w.registryURLInput.SetValue("")
-	}
+	w.registryURLInput.SetValue(agentutils.GetDefaultRegistryURL())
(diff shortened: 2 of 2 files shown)


internal/cli/agent/tui/mcp_wizard.go
```diff
@@ -864,8 +865,12 @@ func (w *McpServerWizard) chooseRemoteType() {
 func (w *McpServerWizard) chooseRegistryType() {
 	w.chosenType = serverTypes.Registry.ID
 	w.step = stepRegistryURL
-	// Clear any stale input value when entering URL step
-	w.registryURLInput.SetValue("")
+	// Pre-populate with the current registry URL if available
+	if defaultURL := agentutils.GetDefaultRegistryURL(); defaultURL != "" {
+		w.registryURLInput.SetValue(defaultURL)
+	} else {
```
peterj: I don't think we need this if/else here necessarily. just setting it to the default registry URL should be fine -- even if the defaultURL is "", we'd be setting it to "" again.

internal/cli/agent/add-mcp.go
```diff
@@ -73,7 +74,7 @@ func addMcpCmd(name string) error {
 
 	// If flags provided, build non-interactively; else run wizard
 	var res models.McpServerType
-	if remoteURL != "" || command != "" || image != "" || build != "" || registryURL != "" || registryServerName != "" { //nolint:nestif
```
peterj, line 76: reason for removing //nolint ? is the linter not complaining anymore?

internal/cli/agent/tui/mcp_wizard.go
optimus-fulcria: Fixed - simplified to just `w.registryURLInput.SetValue(agentutils.GetDefaultRegistryURL())` without the if/else.

internal/cli/agent/add-mcp.go
optimus-fulcria: Restored the `//nolint:nestif` directive - it was accidentally removed in the initial commit.

PR #292, opened 2026-03-06
fix: check --env flags when validating API keys during deploy

### LINKED ISSUE #151: arctl agent deploy ignores the --env setting for the api key
e.g. 

```
 arctl agent deploy myagent  --runtime kubernetes --env GOOGLE_API_KEY=somekey --namespace default
```

fails with:

```
Error: required API key GOOGLE_API_KEY not set for model provider gemini
```

We're checking the environment, but we should check the `--env` flags too and make sure those are used when deploying.

optimus-fulcria:
# Description

- Adds `--env` / `-e` flag to `arctl agent deploy` for passing environment variables (KEY=VALUE)
- `validateAPIKey()` now checks `--env` values before falling back to OS environment
- `buildDeployConfig()` includes all `--env` entries in the deployment config, with flag values taking precedence over OS env
- Fixes the reported issue where `arctl agent deploy myagent --env GOOGLE_API_KEY=somekey` failed with "required API key not set"
- Adds `--env` / `-e` flag to `arctl agent run` for passing environment variables (KEY=VALUE) (for command consistency)
- Updated `agent run` tests to verify behavior.

Fixes #151

```
/kind fix
```

```release-note
Add missing `--env` flag for `arctl agent deploy` to allow setting environment variables on the agent deployment.
```

## Validation Steps

```sh
# assuming you don't have GOOGLE_API_KEY env set already...

# agent deploy test
go run cmd/cli/main.go agent build dice
go run cmd/cli/main.go agent publish dice
go run cmd/cli/main.go agent deploy dice # hit error
go run cmd/cli/main.go agent deploy dice --env GOOGLE_API_KEY=test # works

# agent run test
go run cmd/cli/main.go agent build dice
go run cmd/cli/main.go agent run dice # hit error
go run cmd/cli/main.go agent run dice --env GOOGLE_API_KEY=test # pass
```

## Test plan
- [x] New `TestValidateAPIKey_WithExtraEnv` (6 cases): extra env, OS env, missing, precedence, nil fallback, unknown provider
- [x] New `TestBuildDeployConfig_WithEnvOverrides` (5 cases): override inclusion, precedence, OS fallback, telemetry, absent keys
- [x] All existing agent CLI tests pass


internal/cli/agent/deploy.go
```diff
@@ -131,4 +150,5 @@ func init() {
 	DeployCmd.Flags().String("provider-id", "", "Deployment target provider ID (defaults to local when omitted)")
 	DeployCmd.Flags().Bool("prefer-remote", false, "Prefer using a remote source when available")
 	DeployCmd.Flags().String("namespace", "", "Kubernetes namespace for agent deployment (defaults to current kubeconfig context)")
+	DeployCmd.Flags().StringArrayP("env", "e", []string{}, "Environment variables to set (KEY=VALUE)")
```
inFocus7: nit: could we be more explicit about how these env vars are set/used (set on deployed agent instance)

internal/cli/agent/deploy_test.go
```diff
@@ -0,0 +1,182 @@
...
+			wantErr:       true,
+			errContain:    "ANTHROPIC_API_KEY",
+		},
+		{
+			name:          "extra env takes precedence",
+			modelProvider: "gemini",
+			osEnv:         map[string]string{},
+			extraEnv:      map[string]string{"GOOGLE_API_KEY": "from-flag"},
+			wantErr:       false,
+		},
```
inFocus7: i don't think this test makes sense here since we're not checking precedence/values. i think there's already a test the in test set below this one, so we're good with that one 👍🏼

internal/cli/agent/deploy_test.go
```diff
@@ -0,0 +1,182 @@
...
+			wantKeys:     map[string]string{"OPENAI_API_KEY": "from-os"},
+		},
+		{
+			name: "telemetry endpoint included",
+			manifest: &models.AgentManifest{
+				ModelProvider:     "openai",
+				TelemetryEndpoint: "http://otel:4317",
+			},
+			envOverrides: map[string]string{"OPENAI_API_KEY": "key"},
+			wantKeys:     map[string]string{"OTEL_EXPORTER_OTLP_TRACES_ENDPOINT": "http://otel:4317"},
```
inFocus7: nit suggestion, just to ensure everything is added (i tested locally and it works)

```suggestion
			wantKeys:     map[string]string{"OTEL_EXPORTER_OTLP_TRACES_ENDPOINT": "http://otel:4317", "OPENAI_API_KEY": "key"},
```

inFocus7 (commented):
manually validated (just small test that it works), unit tests are good for sanity too, just had very minor nitpicks

internal/cli/agent/deploy.go
optimus-fulcria: Addressed in 2fdda63 -- updated the flag description to clarify these env vars are set on the deployed agent.

internal/cli/agent/deploy_test.go
optimus-fulcria: Good call, removed in 2fdda63. The "env override takes precedence over os env" test in TestBuildDeployConfig_WithEnvOverrides already covers precedence properly.

internal/cli/agent/deploy_test.go
optimus-fulcria: Applied in 2fdda63 -- wantKeys now includes both OTEL_EXPORTER_OTLP_TRACES_ENDPOINT and OPENAI_API_KEY to verify all env vars are present in the config.

internal/cli/agent/deploy.go
```diff
@@ -44,6 +46,16 @@ func runDeploy(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("API client not initialized")
 	}
 
+	// Parse --env flags into a map
+	envMap := make(map[string]string)
+	for _, e := range envFlags {
```
peterj: @inFocus7 there's a `parseEnvFlags` in `agent/run.go` -- would it make sense to pull that out into some util/common file and then re-use it here too?

inFocus7: Yeah that makes sense. I moved it to a new `internal/cli/utils/utils.go` file, since this is specific for cli usage. If we think it'll be used elsewhere, i can move it to the pre-existing general `internal/utils/utils.go` pkg.

PR #267, opened 2026-03-05
fix: hide registry flags from mcp init help output

optimus-fulcria:
# Description

Hide registry-specific flags (`--registry-url`, `--project-dir`) from the `mcp init` help output since they are not relevant to MCP initialization.

**What changed:**
- Updated `hideRegistryFlags` to accept variadic commands
- Applied to `mcp init` command to hide confusing flags from help text

Fixes #260

# Change Type

/kind fix

# Changelog

```release-note
Hide registry flags from mcp init help output
```

# Additional Notes

The flags still function if explicitly passed — they are only hidden from `--help` output.

PR #293, opened 2026-03-06
fix: hide registry flags from init command help output

### LINKED ISSUE #115: arctl mcp init contains flags it shouldn't

optimus-fulcria:
## Summary
- Hides `--registry-url` and `--registry-token` from `--help` output for `mcp init`, `agent init`, and `skill init` commands
- These init commands are local scaffolding operations that don't interact with the registry, so showing registry flags is confusing
- Uses a custom help function that temporarily hides inherited flags during rendering, then restores them so other commands still show them

Fixes #115

## Test plan
- [x] New `TestInitCmd_HidesRegistryFlags`: verifies registry flags are absent from `mcp init --help` output
- [x] All existing CLI tests pass
- [x] Manually verified: flags still visible for commands that need them (e.g., `mcp list --help`)

### DIFF
--- a/internal/cli/agent/init.go
+++ b/internal/cli/agent/init.go
@@ -11,6 +11,7 @@ import (
 	"github.com/agentregistry-dev/agentregistry/internal/version"
 	"github.com/agentregistry-dev/agentregistry/pkg/validators"
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 )
 
 const adkBaseImageVersion = "0.7.12"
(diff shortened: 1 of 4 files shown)


optimus-fulcria:
Closing - overlaps with my earlier #267 which covers the same issue (#115). Will extend #267 to cover all init commands if needed.

PR #294, opened 2026-03-06
fix: add --tag flag to mcp publish as alias for --version

### LINKED ISSUE #148: the --tag in arctl mcp publish doesn't work
the --version does.

optimus-fulcria:
# Description

**Motivation:** The `arctl skill publish` command has a `--tag` flag, but `arctl mcp publish` does not. This inconsistency means users who try `--tag` with `mcp publish` get a silent failure.

**What changed:** Added `--tag` flag to `arctl mcp publish` as an alias for `--version`. When `--version` is not set, `--tag` is used as the server version. `--version` takes precedence when both flags are provided.

**Related issues:** Fixes #148

/kind fix

# Changelog

```release-note
Add --tag flag to mcp publish command as alias for --version
```

# Additional Notes

- Follows the same precedence pattern as `arctl skill publish`
- New table-driven tests cover: version precedence, tag fallback, both empty, version-only

### DIFF
--- a/internal/cli/mcp/publish.go
+++ b/internal/cli/mcp/publish.go
@@ -28,6 +28,7 @@ var (
 	dryRunFlag          bool
 	overwriteFlag       bool
 	publishVersion      string
+	publishTag          string
 	githubRepository    string
 	publishTransport    string
 	publishTransportURL string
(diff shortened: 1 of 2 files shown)


peterj (changes requested):
thanks. please update the PR template

PR #245, opened 2026-03-03
Adding a Helm Chart to Agent Registry

frivas-at-navteca:
# Description

We are testing and exploring Agent Registry and we have realized that deploying it to Kubernetes is easier using a Helm Chart, we could not find one and that is why we build it.

This will be my first contribution to the project, please bear with me, as adding a license is needed, following the project's guidelines as well, etc. I would say this is a bare bones Helm Chart. I am completely aware that the team needs to review it and most likely I will need to make changes, happy to do so.

**What is included:**
**- Application Management:** Exposing and HTTP API through port 12121, I can access the UI in my browser through this port using for example, `kubectl port-forward` command. Also a gRPC Agent Gateway through 21212.
**- Includes a PostgreSQL with pgvector:** One can choose the version to be deployed. Persistent storage through StorageClass. It also supports an external database if you don't want the bundled one.
**- Security:** Non-root containers with read-only root filesystem, secrets are auto-generated on first install. Hex-encoded JWT. Supports pre-existing secrets so you can integrate with external secret managers. Network policies with granular rules. Least-privilege RBAC by default.
**- Kubernetes Operations:** liveness, readiness and startup health probes. HPA with CPU and Memory targets. PDB for voluntary disruptions. Resource presets and affinity presets or full override block for more control if needed.
**- Extensions:** You can change and extend through extraEnvVars, extraVolumes; commonLabels and commonAnnotations are also supported for better tagging. Global overrides for image, pull secrets, StorageClass etc.

I can also provide a minimal and basic terraform code to deploy this chart.

I was about to open an Issue and found #242 this morning so I realized that at least there is one more person interested on having a Helm Chart for the Agent Registry so I decided to go ahead and open a PR. I hope this is ok.

# Change Type

```
/kind feature
```

# Release Notes

```release-note
add Helm chart to agentregistry
```

# Additional Notes

I am happy to maintain this Chart for the project.

**I have a few questions/comments to the team:**

1. What license should I use for the Chart?
2. Where shall it be hosted? what should be the release cycle?
3. I have followed Bitnami's Helm Charts best practices. I have been using those Charts for a couple years and I like their approach however if the project team is not happy with that, I can make the changes needed. 

Please, make suggestions and comments so this Helm can be improved and hopefully approved by the team. Again, bear with me, this is my first contribution.

Issue #242, opened 2026-03-02
Add helm charts


timflannagan:
I'd like an easy way to deploy the control plane on a kube/kind environment. Having helm charts that allow me to deploy an env would help quite a bit when testing out new features using the --provider-id=kubernetes-default (or whatever that provider name is called).

frivas-at-navteca:
I have built one for Agent Registry. I was about to open a discussion about it. I would like to get in touch with the project team to get the details on how to proceed. 

I have already built it and tested it in kind and minikube I have even used terraform to deploy it. 

My questions/comments to the team are:
1. Where would you like to host it?
2. I can be the maintainer of it.
3. What is the procedure to review it and make it official?

I am happy to contribute.

frivas-at-navteca:
@timflannagan Thank you very much! I am wondering how to proceed now with the PR #245  given the workflow is mostly created for a Go based code of course.

timflannagan:
@frivas-at-navteca Just posted a quick review on that PR. Thank you!

frivas-at-navteca:
Thank you very much @timflannagan I truly appreciate your time and comments. I (believe) have addressed them all .

frivas-at-navteca:
Thank you very much @timflannagan @nikolasmatt and @peterj for your patience and for simplifying the chart, awesome!.

I still have one thing I am wondering about: what happens with kMCP? in our tests to properly deploy an MCP, meaning actually deploying the pod kMCP (CRDs and the app itself) seems to be needed. Is my understanding correct? Shall I create a new issue asking about this?

Thank you very much again! I saw the readme updated. Very cool!

timflannagan:
Closing this out now that we have an initial helm chart hooked up.

@frivas-at-navteca Good Q. Less familiar with the state of the kmcp project or whether kagent replicates most of the functionality today. You can achieve similar results by deploying an MCP server through our stack, which uses kagent CRDs under the hood, whenever you use deploy to Kubernetes environments. Let's open a new issue if needed.

PR #297, opened 2026-03-06
feature(helm): charts followup

frivas-at-navteca:
# Description
This is a follow up to PR #245. This PR addresses the suggestions from @peterj and @timflannagan. This continues the implementation of #242. This PR includes 2 new workflows one to test and another to create a release of the chart. The set of targets in the Makefile just for the charts. I have tried to be very careful on messing with something else.

**I want to bring to your attention a couple things:**
1. There are variables that need to be set for the workflow to work (I can change them if needed):

```
HELM_REGISTRY_PASSWORD
HELM_REGISTRY
HELM_REGISTRY_USERNAME
HELM_REPO
```

2. In our tests we have noticed that you always need kMCP (CRDs and the tool itself). We initially through it was kAgent but we narrowed it down to kMCP so Agent Registry can actually created the pod. We tested with just the CRDs, the MCP is deployed but no pod is created, with the tool itself it seems to work correctly.

Shall I add kMCP as a dependency of the Agent Registry chart and also create a section in the values.yaml in which it can be enabled and configured? otherwise administrators will have to do this separately and it seems the documentation is not including this information.

# Change type
/kind cleanup

# Changelog
```release-note
Adding a Helm Chart
```

### DIFF
--- a/.github/workflows/chart-tests.yml
+++ b/.github/workflows/chart-tests.yml
@@ -6,14 +6,12 @@ on:
     paths:
       - 'charts/**'
       - '.github/workflows/chart-tests.yml'
-      - '.github/workflows/chart-release.yml'
       - 'Makefile'
   pull_request:
     branches: [ main ]
     paths:
       - 'charts/**'
       - '.github/workflows/chart-tests.yml'
-      - '.github/workflows/chart-release.yml'
       - 'Makefile'
   merge_group:
     branches: [ main ]


Makefile
```diff
@@ -276,3 +298,89 @@ mod-tidy: ## Run go mod tidy
...
+	    || { echo "ERROR: helm-unittest install failed. Check network / plugin URL."; exit 1; }; \
+	else \
+	  echo "helm-unittest plugin already installed"; \
+	fi
+	@echo "Running helm-unittest on $(HELM_CHART_DIR)..."
+	$(HELM) unittest $(HELM_CHART_DIR) --file "tests/*_test.yaml"
+
+# Convenience: package → push → test.
+.PHONY: charts-all
+charts-all: charts-push charts-test
```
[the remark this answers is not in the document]
frivas-at-navteca: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/agentregistry-dev/agentregistry/pull/297#discussion_r2896055474)

.github/workflows/chart-release.yml
```diff
@@ -0,0 +1,205 @@
...
+
+          if [ "${{ steps.creds.outputs.password_source }}" = "secret" ]; then
+            PASSWORD="${{ secrets.HELM_REGISTRY_PASSWORD }}"
+          else
+            PASSWORD="${{ secrets.GITHUB_TOKEN }}"
+          fi
+
+          printf "%s" "${PASSWORD}" | helm registry login "${REGISTRY}" \
+            --username "${USERNAME}" \
+            --password-stdin
```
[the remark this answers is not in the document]
frivas-at-navteca: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/agentregistry-dev/agentregistry/pull/297#discussion_r2896055488)

Makefile
```diff
@@ -276,3 +298,89 @@ mod-tidy: ## Run go mod tidy
...
+#   1. checks that 'helm' is present (fails with a clear message if not)
+#   2. checks for the helm-unittest plugin and installs it if missing
+#   3. runs the full test suite
+.PHONY: charts-test
+charts-test: _helm-check charts-deps
+	@echo "Checking for helm-unittest plugin..."
+	@if ! $(HELM) plugin list | awk '{print $$1}' | grep -q '^unittest$$'; then \
+	  echo "helm-unittest plugin not found — installing from $(HELM_PLUGIN_UNITTEST_URL)"; \
+	  $(HELM) plugin install $(HELM_PLUGIN_UNITTEST_URL) --version $(HELM_PLUGIN_UNITTEST_VERSION) $(HELM_PLUGIN_INSTALL_FLAGS) \
+	    || { echo "ERROR: helm-unittest install failed. Check network / plugin URL."; exit 1; }; \
```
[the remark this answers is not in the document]
frivas-at-navteca: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/agentregistry-dev/agentregistry/pull/297#discussion_r2896055599)

.github/workflows/chart-release.yml
```diff
@@ -0,0 +1,205 @@
...
+      # ── Push to GHCR (OCI) ───────────────────────────────────────────────────
+      - name: Resolve registry credentials
+        id: creds
+        run: |
+          # Use explicit secret if provided; fall back to built-in GITHUB_TOKEN.
+          if [ -n "${{ secrets.HELM_REGISTRY_PASSWORD }}" ]; then
+            echo "password_source=secret" >> "$GITHUB_OUTPUT"
+          else
+            echo "password_source=github_token" >> "$GITHUB_OUTPUT"
+          fi
```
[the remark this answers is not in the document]
frivas-at-navteca: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/agentregistry-dev/agentregistry/pull/297#discussion_r2896055540)

.github/workflows/chart-release.yml
```diff
@@ -0,0 +1,229 @@
...
+# Optional repository secrets (defaults shown):
+#   HELM_REGISTRY           – default: ghcr.io
+#   HELM_REGISTRY_USERNAME  – default: github.actor
+#
+# Consumers can pull the published chart with:
+#   helm pull oci://ghcr.io/agentregistry-dev/agentregistry/charts/agentregistry --version <ver>
+
+on:
+  release:
+    types: [ published ]
```
nikolasmatt: the `release.yml` workflow builds the images and then creates the release.
This reacts to the release and attaches the charts after the fact.
If for some reason this fails, it will cause a half built release.

Please merge this with the release workflow so all the artifacts before the release.

.github/workflows/chart-release.yml
```diff
@@ -0,0 +1,229 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up Helm
+        uses: azure/setup-helm@v4
+        with:
+          version: '3.17.0'
```
nikolasmatt: ubunute-latest comes with helm `3.20` [pre-installed](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#package-management). Do we need to downgrade?

.github/workflows/chart-release.yml
```diff
@@ -0,0 +1,229 @@
...
+
+      - name: Cache Helm plugins
+        uses: actions/cache@v4
+        with:
+          path: ~/.local/share/helm/plugins
+          key: ${{ runner.os }}-helm-plugins-unittest-${{ hashFiles('charts/agentregistry/Chart.yaml') }}
```
nikolasmatt: helm plugins are not keyed off of the state of `Chart.yaml`. I'd rather this caches the unittest plugin directly for now and key the cache off of the plugins version

.github/workflows/chart-release.yml
```diff
@@ -0,0 +1,229 @@
...
+            ${{ runner.os }}-helm-plugins-unittest-
+
+      - name: Install helm-unittest plugin (if missing)
+        run: |
+          set -euo pipefail
+          if helm plugin list | awk '{print $1}' | grep -q '^unittest$'; then
+            echo "helm-unittest already installed"
+          else
+            helm plugin install https://github.com/helm-unittest/helm-unittest --version v1.0.3
+          fi
```
nikolasmatt: you have this logic duplicated in the make file. just split it to it's own make target and reuse that

.github/workflows/chart-release.yml
```diff
@@ -0,0 +1,229 @@
...
+      - name: Build chart dependencies
+        run: helm dependency build charts/agentregistry
+
+      - name: Lint chart (strict)
+        run: helm lint charts/agentregistry --strict
+
+      - name: Run helm-unittest
+        run: |
+          helm unittest charts/agentregistry \
+            --file "tests/*_test.yaml"
```
nikolasmatt: these should use make targets instead so they can be easily executed manually locally

.github/workflows/chart-tests.yml
```diff
@@ -0,0 +1,99 @@
+name: Chart Tests
+
+on:
```
nikolasmatt, line 3: this is missing this trigger
```yaml
  merge_group:
    branches: [main]
    types: [checks_requested]
```

.github/workflows/chart-tests.yml
```diff
@@ -0,0 +1,99 @@
+name: Chart Tests
```
nikolasmatt, line 1: review this workflow after creating the make targets i noted on the release workflow.

charts/agentregistry/values.yaml
```diff
@@ -381,25 +381,26 @@ database:
...
     image:
       # -- PostgreSQL image registry
       registry: docker.io
       # -- PostgreSQL image repository (pgvector/pgvector)
       repository: pgvector/pgvector
-      # -- PostgreSQL image tag — MUST use immutable tags (not `pg17` or `latest`)
-      tag: "0.8.2-pg17"
+      # -- PostgreSQL image tag — immutable tags are recommended for reproducible builds;
+      #    using floating tags (e.g. `pg16`, `pg17`, `latest`) may lead to non-reproducible deployments
+      tag: "pg16"
```
nikolasmatt: @peterj you asked for this change in the previous PR but the comment is valid here. Maybe we should use `0.8.2-pg16` in both the chart and the docker setup?

Makefile
```diff
@@ -23,7 +23,21 @@ LDFLAGS := \
...
+HELM_CHART_DIR ?= ./charts/agentregistry
+HELM_PACKAGE_DIR ?= build/charts
+HELM_REGISTRY ?= ghcr.io
+HELM_REPO ?= agentregistry-dev/agentregistry
+# HELM_PUSH_MODE: oci (default, recommended) | repo (legacy chart repo / ChartMuseum)
+HELM_PUSH_MODE ?= oci
+HELM_PLUGIN_UNITTEST_URL ?= https://github.com/helm-unittest/helm-unittest
+# Pin the helm-unittest plugin version for reproducibility and allow install flags
+HELM_PLUGIN_UNITTEST_VERSION ?= v1.0.3
+HELM_PLUGIN_INSTALL_FLAGS ?= --verify=false
```
nikolasmatt, line 43: why are we skipping verify?

Makefile
```diff
@@ -23,7 +23,21 @@ LDFLAGS := \
...
+
+.PHONY: help install-ui build-ui clean-ui build-cli build install dev-ui run down test test-integration test-coverage test-coverage-report clean lint all release-cli docker-compose-up docker-compose-down docker-compose-logs charts-deps charts-lint charts-package charts-push charts-test charts-all
```
nikolasmatt: `.PHONY` only needs to be defined once for each target

frivas-at-navteca:
> Thanks for the contribution and the rapid iterations on this work! I left a few comments for improvements primarily themed around release coordination and code-reuse between workflows and local.

Thank you very. I appreciate the time and your comments. I will address them.

.github/workflows/chart-release.yml
frivas-at-navteca: Sure. Good catch.

.github/workflows/chart-release.yml
frivas-at-navteca: Definitely, I changed it to use that one instead.

.github/workflows/chart-release.yml
frivas-at-navteca: Understood. I changed it.

.github/workflows/chart-release.yml
frivas-at-navteca: Yes. Also changed it.

.github/workflows/chart-release.yml
frivas-at-navteca: Sure. Done.

.github/workflows/chart-tests.yml
frivas-at-navteca: Added.

Makefile
frivas-at-navteca: Just to confirm my understanding. There should be one .PHONY for each target instead of that big .PHONY correct?

Makefile
frivas-at-navteca: In my tests --verify was failing

```
Run set -euo pipefail
  set -euo pipefail
  if helm plugin list | awk '{print $1}' | grep -q '^unittest$'; then
    echo "helm-unittest already installed (from cache or prior install)"
  else
    echo "Installing helm-unittest plugin..."
    helm plugin install https://github.com/helm-unittest/helm-unittest --version v1.0.3 --verify=false
  fi
  shell: /usr/bin/bash -e {0}
Installing helm-unittest plugin...
Error: unknown flag: --verify
Error: Process completed with exit code 1.
```

It seems because the plugin does not support the verification. I will look into this.

I am seeing these 
https://github.com/helm/helm/issues/31490
https://github.com/Azure/setup-helm/issues/239

Which are the errors I have seen while testing. I will double check.

Makefile
nikolasmatt: it only needs to be on one or the other. Personally i prefer defining it next to the target but either works

Makefile
nikolasmatt: good to know! do you mind adding a comment next to it with those links for future reference?

charts/agentregistry/values.yaml
nikolasmatt: ~i'm resolving this for now (lies i can't resolve issues)~ cause there is some follow up work that will either address it or eliminate it's need. It's ok to leave it as is in this PR.

frivas-at-navteca: Ok. I will leave it as the immutable recommended tag.

Makefile
frivas-at-navteca: Sure. Added.

Makefile
frivas-at-navteca: Ok. I have added one .PHONY per target.

nikolasmatt (approved):

Issue #312, opened 2026-03-11
consolidate appVersion and version in Chart.yaml


peterj:
Is there a way we could manually update the versions there too based on the pushed tag?

nikolasmatt:
I can take care of this to update automatically

PR #320, opened 2026-03-12
Automatically update `Chart.yaml` version for release

nikolasmatt:
# Description
Fixes https://github.com/agentregistry-dev/agentregistry/issues/312

Removes the need to manually update `Chart.yaml` with the release version.
Moves most of the helmchart release logic from the GH workflow file to make targets

# Change Type

# Changelog
/kind cleanup

```release-note
NONE
```

# Additional Notes

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -103,7 +103,7 @@ jobs:
       run: make release-cli
 
     - name: Release Helm chart
-      run: make release-helm
+      run: make charts-release
 
     - name: Create GitHub Release
       uses: softprops/action-gh-release@v2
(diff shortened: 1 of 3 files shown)


charts/agentregistry/Chart-template.yaml
```diff
@@ -26,6 +25,4 @@ annotations:
       url: https://github.com/agentregistry-dev/agentregistry
   artifacthub.io/images: |
     - name: agentregistry
-      image: ghcr.io/agentregistry-dev/agentregistry/server:0.2.1
-    - name: pgvector
-      image: pgvector/pgvector:0.8.2-pg16
```
peterj, line 31: was deleting the pgvector here intentional?

nikolasmatt: the chart no longer installs postgres, so the chart shouldnt depend on pgvector any more, right?

peterj: yep, all good. I thought it was removed already

PR #346, opened 2026-03-13
internal: Fix failed deployment cleanup

### LINKED ISSUE #326: Failed deployment records cannot be deleted
Related to #322. I have an environment where there's an agent deployment in a failed state. I tried to hit the registry API directly in order to unblock the environment, but ran into the following 500 response:

```bash
$ kubectl -n agentregistry exec deploy/agentregistry -- \
  curl -sS -X DELETE \                                                                                                                                             http://127.0.0.1:8080/v0/deployments/4c3e848f-d98b-4cd3-8698-e06abb9b6ebf | jq
{
  "title": "Internal Server Error",
  "status": 500,
  "detail": "Failed to remove deployment",
  "errors": [
    {
      "message": "failed to list agents by deployment id 4c3e848f-d98b-4cd3-8698-e06abb9b6ebf: agents.kagent.dev is forbidden: User \"system:serviceaccount:agentregistry:agentregistry\" cannot list resource \"agents\" in API group \"kagent.dev\" in the namespace \"agentregistry\""
    }
  ]
}
```

In this scenario, the deployment is already in a failed state. Expected behavior: failed or cancelled deployment records should be removable without requiring runtime undeploy?

timflannagan:
# Description

We stop treating failed or cancelled deployment rows as active deployments and allow those rows to be removed without invoking platform undeploy.

Previously, list and metadata surfaces counted any matching deployment row, including failed and cancelled records. Deleting a failed deployment also still tried to undeploy platform resources first, which could fail even when nothing had been created.

Now, only deployed rows contribute to deployment status displays, and failed or cancelled records are removed directly from the database during delete.

Fixes #326.

# Change Type

/kind fix

# Changelog

```release-note
NONE
```

# Additional Notes

### DIFF
--- a/internal/cli/common/deployments.go
+++ b/internal/cli/common/deployments.go
@@ -4,13 +4,18 @@ import (
 	"fmt"
 
 	"github.com/agentregistry-dev/agentregistry/internal/client"
+	"github.com/agentregistry-dev/agentregistry/pkg/models"
 )
 
 // BuildDeploymentCounts indexes deployment rows by resource name and version.
+// Skip non-deployed statuses and non-matching resource types.
 func BuildDeploymentCounts(deployments []*client.DeploymentResponse, resourceType string) map[string]map[string]int {
 	counts := make(map[string]map[string]int)
 	for _, deployment := range deployments {
 		if deployment == nil || deployment.ResourceType != resourceType {
+			continue
+		}
+		if deployment.Status != models.DeploymentStatusDeployed {
 			continue
 		}
 		if counts[deployment.ServerName] == nil {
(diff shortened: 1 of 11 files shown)


internal/cli/common/deployments.go
```diff
@@ -4,15 +4,20 @@ import (
...
 
 // BuildDeploymentCounts indexes deployment rows by resource name and version.
+// Skip non-deployed statuses and non-matching resource types.
 func BuildDeploymentCounts(deployments []*client.DeploymentResponse, resourceType string) map[string]map[string]int {
 	counts := make(map[string]map[string]int)
 	for _, deployment := range deployments {
 		if deployment == nil || deployment.ResourceType != resourceType {
 			continue
 		}
+		if deployment.Status != models.DeploymentStatusDeployed {
```
timflannagan, line 18: TODO: confirm whether we care about the legacy (afaik) "active" column. There was internal/registry/database/migrations/004_unified_deployments_provider.sql and I thought we ripped this out in Scott's refactor a couple of weeks back.

timflannagan: cc @peterj @ilackarms

PR #347, opened 2026-03-13
WIP: Wait for Kubernetes readiness

### LINKED ISSUE #230: Deploying an agent to k8s reports successful state even when Agent isn't ready
When I deploy an agent to the kubernetes runtime that has kagent installed, I see the underlying kagent `Agent` CR in Ready=false with DeploymentNotReady state. The CLI itself marks the operation as successful. I'd expect as a user that the CLI waits until any runtime resource (i.e. the Agent in this scenario) is in a ready state before marking the command run as successful. Maybe add a --wait=false CLI flag.

timflannagan:
# Description

We now keep Kubernetes deployments in an in-flight state until the live platform resources report readiness, and the CLI wait path surfaces current condition details instead of treating apply as success.

Previously, the Kubernetes adapter returned deployed immediately after apply, so --wait could exit successfully while the underlying Agent or MCP resources were still not ready. Follow-up to #296 which introduced the initial scaffolding, but didn't actually work.

Now, Kubernetes deploys stay in deploying until GetDeploymentByID refreshes their state from live CR conditions, and wait failures include the current deployment error text.

Fixes #230.

# Change Type

/kind fix

# Changelog

```release-note
NONE
```

# Additional Notes

### DIFF
--- a/internal/cli/common/wait.go
+++ b/internal/cli/common/wait.go
@@ -2,12 +2,13 @@ package common
 
 import (
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/agentregistry-dev/agentregistry/internal/client"
 )
 
-const (
+var (
 	defaultWaitTimeout  = 5 * time.Minute
 	defaultPollInterval = 2 * time.Second
 )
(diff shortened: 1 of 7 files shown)


timflannagan:
Worked locally but mostly AI generated. Want to manually validate these changes. Adding WIP label for now.

PR #363, opened 2026-03-17
add back deploy functionality to UI

### LINKED ISSUE #352: deploying from the UI
when we removed the "published" concept, we also removed the "deploy" functionality from the UI (because it was in the published UI). 

Do we want to bring the ability to deploy from the UI back? If so, we'll need to figure out how do we support this. If the backend/ui is running in a docker container, we can deploy both to Docker and Kubernetes. Howevwr, if we're running inside the k8s cluster we don't have access to the local environment and can't deploy to docker. 

To implement this properly, we'll probably need to have something that will allow us to check which providers are avialable and/or we can deploy to. Once ew have that, we can show the options in the UI (we have the REST api there anyway) .

peterj:
# Description

Adds back the Deploy functionality that was removed when we removed the publish page.

fixes #352 

# Change Type

```
/kind feature
```

# Changelog

```release-note
add the Deploy UI
```

### DIFF
--- a/ui/lib/api/types.gen.ts
+++ b/ui/lib/api/types.gen.ts
@@ -282,9 +282,9 @@ export type HealthBody = {
      */
     github_client_id?: string;
     /**
-     * Platform mode: docker or kubernetes
+     * Platform mode
      */
-    platform_mode?: string;
+    platform_mode?: 'docker' | 'kubernetes';
     /**
      * Health status
      */


internal/registry/api/handlers/v0/health.go
```diff
@@ -18,6 +18,7 @@ import (
 type HealthBody struct {
 	Status         string `json:"status" example:"ok" doc:"Health status"`
 	GitHubClientID string `json:"github_client_id,omitempty" doc:"GitHub OAuth App Client ID"`
+	PlatformMode   string `json:"platform_mode,omitempty" example:"docker" doc:"Platform mode: docker or kubernetes"`
```
inFocus7: Could use an `enum` to enforce this
```suggestion
	PlatformMode   string `json:"platform_mode,omitempty" example:"docker" doc:"Platform mode" enum:"docker,kubernetes"`
```

ui/components/server-card.tsx
```diff
@@ -127,24 +128,44 @@ export function ServerCard({ server, onDelete, onDeploy, showDelete = false, sho
...
+                size="sm"
+                className="h-7 gap-1 text-xs"
+                onClick={(e) => { e.stopPropagation(); onDeploy(server) }}
+              >
+                <Play className="h-3 w-3" aria-hidden="true" />
+                Deploy
+              </Button>
+            ) : (
+              <Tooltip>
+                <TooltipTrigger asChild>
```
inFocus7, line 143: q: how does the tooltip work? i published a non-oci server and the deploy button is disabled as expected, but I'm not seeing the tooltip on hover 👀

peterj: fixed:

Screenshot 2026-03-18 at 2 30 31 PM

Issue #358, opened 2026-03-17
Bundle postgres inside the agentregistry chart by default


pmuir:
Currently we tell people
> Deploy a single-instance PostgreSQL and pgvector into your cluster using the provided example manifest:
> 
> kubectl apply -f https://raw.githubusercontent.com/agentregistry-dev/agentregistry/main/examples/postgres-pgvector.yaml
> kubectl -n agentregistry wait --for=condition=ready pod -l app=postgres-pgvector --timeout=120s
> This setup is intended for development and testing. For production, use a managed PostgreSQL service or a production-grade operator.

Instead, by default, we should include postgres inside the main agentregistry chart, with an option to disable it and configure your own.

PR #365, opened 2026-03-18
Reintroduce bundled database to the chart

nikolasmatt:
Resolves https://github.com/agentregistry-dev/agentregistry/issues/358

# Description

Reintroduces a bundled PostgreSQL instance to the Helm chart and restructures the `database` values key to align with [kagent PR #1527](https://github.com/kagent-dev/kagent/pull/1527).

**What changed:**

- **Bundled PostgreSQL on by default** — `postgres:18` is deployed alongside Agent Registry when `database.postgres.bundled.enabled=true` (the default). Intended for development and evaluation only; a warning is shown in NOTES.txt.
- **Restructured `database` values** — replaced the old `database.bundled.*` / `database.external.*` split with a flat `database.postgres.*` structure:
  - `database.postgres.url` / `database.postgres.urlFile` — external connection string (urlFile takes precedence)
  - `database.postgres.bundled.*` — bundled instance config (image, storage, resources)
  - `database.postgres.vectorEnabled` — enables `AGENT_REGISTRY_EMBEDDINGS_ENABLED` and `AGENT_REGISTRY_DATABASE_VECTOR_ENABLED` on the app pod
- **Default image is `postgres:18`** (official, plain). Users who want semantic search can opt in via `database.postgres.vectorEnabled=true` with their own pgvector-capable PostgreSQL instance.
- **Conditional vector migrations** — base schema no longer requires pgvector. Vector schema (embedding columns, HNSW indexes) is applied in a separate migration set, only when `DatabaseVectorEnabled=true`.
- **URL file support** — `AGENT_REGISTRY_DATABASE_URL_FILE` env var / `database.postgres.urlFile` value reads the database URL from a file at startup.
- **Separate PostgreSQL Secret** — bundled DB password lives in its own `{release}-postgresql` Secret, decoupled from the JWT key Secret.
- **Init container only when needed** — the `wait-for-postgres` init container is skipped when `url` or `urlFile` is set, even if `bundled.enabled=true`.
- **Simplified helpers** — removed `agentregistry.secretName`, `agentregistry.passwordSecretName`, and `agentregistry.databaseUrl`; logic inlined into templates. Removed `global.existingSecret`.
- **Updated docs** — `README.md`, `DEVELOPMENT.md`, chart `README.md.gotmpl`, and `NOTES.txt` updated for accuracy.
- **Helm unit tests** — new `postgresql_test.yaml` suite; `deployment_test.yaml`, `secrets_test.yaml`, and `validation_test.yaml` updated for the new values structure.

# Change Type

/kind feature
/kind install

# Changelog

```release-note
Reintroduces a bundled PostgreSQL database option to the Helm chart with a new database.postgres.* values structure. The bundled database (postgres:18) is enabled by default for development and evaluation. Set database.postgres.bundled.enabled=false and supply database.postgres.url for production.
```

### DIFF
--- a/charts/agentregistry/templates/postgresql-secret.yaml
+++ b/charts/agentregistry/templates/postgresql-secret.yaml
@@ -4,8 +4,7 @@ kind: Secret
 metadata:
   name: {{ include "agentregistry.fullname" . }}-postgresql
   labels:
-    {{- include "agentregistry.labels" . | nindent 4 }}
-    app.kubernetes.io/component: database
+    {{- include "agentregistry.postgresql.labels" . | nindent 4 }}
   {{- $annotations := include "agentregistry.annotations" (dict "annotations" dict "context" $) }}
   {{- if $annotations }}
   annotations:
--- a/charts/agentregistry/templates/_helpers.tpl
+++ b/charts/agentregistry/templates/_helpers.tpl
@@ -53,6 +53,7 @@ Selector labels — stable subset used in matchLabels.
 {{- define "agentregistry.selectorLabels" -}}
 app.kubernetes.io/name: {{ include "agentregistry.name" . }}
 app.kubernetes.io/instance: {{ .Release.Name }}
+app.kubernetes.io/component: server
 {{- end }}
 
 {{/*
@@ -270,8 +271,18 @@ Full name for the bundled PostgreSQL resources.
 Standard labels for bundled PostgreSQL resources.
 */}}
 {{- define "agentregistry.postgresql.labels" -}}
-{{ include "agentregistry.labels" . }}
+helm.sh/chart: {{ include "agentregistry.chart" . }}
+app.kubernetes.io/name: {{ include "agentregistry.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
 app.kubernetes.io/component: database
+{{- if .Chart.AppVersion }}
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+{{- end }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+app.kubernetes.io/part-of: {{ include "agentregistry.name" . }}
+{{- if .Values.commonLabels }}
+{{ toYaml .Values.commonLabels }}
+{{- end }}
 {{- end }}
 
 {{/*


README.md
```diff
@@ -108,28 +108,26 @@ Open `http://localhost:12121` to use the web UI.
...
+helm install agentregistry oci://ghcr.io/agentregistry-dev/agentregistry/charts/agentregistry \
+  --namespace agentregistry \
+  --create-namespace \
+  --set config.jwtPrivateKey=$(openssl rand -hex 32)
 ```
 
-This setup is intended for development and testing. For production, use a managed PostgreSQL service or a production-grade operator.
-
-#### Install Agent Registry
+For production, disable the bundled database and point to a managed PostgreSQL service:
```
iplay88keys: I think this could be more strongly called out. Maybe with one of [these](https://github.com/orgs/community/discussions/16925)?

charts/agentregistry/templates/postgresql.yaml
```diff
@@ -0,0 +1,131 @@
...
+{{- $pg := .Values.database.bundled }}
+{{- $fullname := include "agentregistry.postgresql.fullname" . }}
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: {{ $fullname }}
+  labels:
+    {{- include "agentregistry.postgresql.labels" . | nindent 4 }}
+  {{- $annotations := include "agentregistry.annotations" (dict "annotations" dict "context" $) }}
```
iplay88keys, line 11: Since this is duplicated down below multiple times, can we just move it to the top of the file and make it shared?

charts/agentregistry/values.yaml
```diff
@@ -5,7 +5,7 @@ global:
   imageRegistry: ""
   # -- Global Docker registry secret names
   imagePullSecrets: []
-  # -- Name of an existing Secret containing all credentials (POSTGRES_PASSWORD and AGENT_REGISTRY_JWT_PRIVATE_KEY). Overrides config.existingSecret and database.existingSecret when set.
+  # -- Name of an existing Secret containing all credentials (POSTGRES_PASSWORD and AGENT_REGISTRY_JWT_PRIVATE_KEY). Overrides config.existingSecret and database.external.existingSecret when set. When database.bundled.enabled=true, this secret must contain POSTGRES_PASSWORD — the chart will not create its own secret and the bundled PostgreSQL pod will reference it directly.
   existingSecret: ""
```
iplay88keys: I think that it would be better to just drop `global.existingSecret` and if someone wants to put both `POSTGRES_PASSWORD` and `AGENT_REGISTRY_JWT_PRIVATE_KEY` in the same secret, they can just point `db.external.existingSecret` and `config.existingSecret` at that same secret since the keys are different.

charts/agentregistry/values.yaml
```diff
@@ -263,19 +267,59 @@ rbac:
...
+    # -- External database host
+    host: ""
+    # -- External database port
+    port: 5432
+    # -- External database name
+    database: "agent-registry"
+    # -- External database user
+    username: "agentregistry"
+    # -- External database password (ignored when existingSecret is set)
+    password: ""
```
iplay88keys: Have you seen any issues with this if the password contains something like: `@, :, /, ?, #, %` which could break url parsing when loaded?

charts/agentregistry/values.yaml
```diff
@@ -263,19 +267,59 @@ rbac:
...
-  # -- External database password (ignored when existingSecret is set)
-  password: ""
-  # -- Name of an existing Secret containing the database password (key: POSTGRES_PASSWORD)
-  existingSecret: ""
-  # -- External database SSL mode (require, verify-ca, verify-full, disable). Defaults to require for encrypted connections.
-  sslMode: "require"
+  # -- Bundled PostgreSQL (dev/test). Enabled by default; set bundled.enabled=false to bring your own.
+  bundled:
+    # -- Deploy a PostgreSQL instance alongside Agent Registry
+    enabled: true
```
iplay88keys: Can we document why we need to have this? Can we key off of the external database configuration?

iplay88keys: Per a discussion with @nikolasmatt, we are going to use bundled if `database.postgres.url` and `database.postgres.host` are both unset instead of having a `database.postgres.bundled.enabled` field.

charts/agentregistry/templates/postgresql-secret.yaml
```diff
@@ -0,0 +1,17 @@
...
+    {{- include "agentregistry.labels" . | nindent 4 }}
+    app.kubernetes.io/component: database
+  {{- $annotations := include "agentregistry.annotations" (dict "annotations" dict "context" $) }}
+  {{- if $annotations }}
+  annotations:
+    {{- $annotations | nindent 4 }}
+  {{- end }}
+type: Opaque
+data:
+  POSTGRES_PASSWORD: {{ "agentregistry" | b64enc | quote }}
```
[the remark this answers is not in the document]
nikolasmatt: this is intentional for now

internal/registry/database/testutil.go
```diff
@@ -128,12 +128,30 @@ func ensureVectorExtension(ctx context.Context, uri string) error {
 	return nil
 }
 
+type testDBOption struct {
+	vectorEnabled bool
+}
+
+// WithVector enables vector migrations (adds semantic_embedding columns) on the test database.
+// Use for tests that exercise pgvector/embeddings functionality.
+func WithVector() testDBOption {
```
timflannagan, line 139: nit: Typically the functions pattern would uses a function type vs. a struct:

```go
type testDBOption func(*testDBConfig)

type testDBConfig struct {
    vectorEnabled bool
}

func WithVector() testDBOption {
    return func(cfg *testDBConfig) {
        cfg.vectorEnabled = true
    }
 }

...
var cfg testDBConfig
for _, o := range opts { o(&cfg) }
```

scripts/kind/README.md
```diff
@@ -20,21 +20,20 @@ This single command sets up the full local environment.
 
 ## What It Does
 
-`make setup-kind-cluster` runs three sub-targets in order:
+`make setup-kind-cluster` runs two sub-targets in order:
```
timflannagan: Fine for now, but this is something that's a moving goalpost if we ever refactor this Makefile target

charts/agentregistry/values.yaml
```diff
@@ -263,19 +267,36 @@ rbac:
...
-  existingSecret: ""
-  # -- External database SSL mode (require, verify-ca, verify-full, disable). Defaults to require for encrypted connections.
-  sslMode: "require"
+  postgres:
+    # -- External PostgreSQL connection string. Used if set, regardless of bundled.enabled.
+    url: ""
+    # -- Path to a file containing the database URL. Takes precedence over url.
+    urlFile: ""
+    # -- Enable vector schema migrations and embeddings/semantic search (sets AGENT_REGISTRY_DATABASE_VECTOR_ENABLED and AGENT_REGISTRY_EMBEDDINGS_ENABLED). Requires a pgvector-capable PostgreSQL instance.
+    vectorEnabled: false
```
timflannagan: This field living under the postgres section is a bit confusing to me. It's effectively configuring fields that live on the server (vs. configuring the postgres instance) and tells it to run vector migrations on startup + enable semantic search functionality. WYDT?

charts/agentregistry/values.yaml
```diff
@@ -263,19 +267,36 @@ rbac:
...
+        # -- Bundled PostgreSQL image repository (org/namespace)
+        repository: library
+        # -- Bundled PostgreSQL image name
+        name: postgres
+        # -- Bundled PostgreSQL image tag
+        tag: "18"
+        # -- Bundled PostgreSQL image pull policy
+        pullPolicy: IfNotPresent
+      # -- DB name, user, and password are hardcoded ("agentregistry") for the bundled instance.
+      # -- PersistentVolumeClaim size for the bundled PostgreSQL data directory
```
timflannagan, line 300: Do we need to expose the ability toggle the class name too? Right now it uses the default class in the cluster, which is probably sufficient for now.

charts/agentregistry/values.yaml
```diff
@@ -263,19 +267,36 @@ rbac:
...
-  # -- External database user
-  username: "agentregistry"
-  # -- External database password (ignored when existingSecret is set)
-  password: ""
-  # -- Name of an existing Secret containing the database password (key: POSTGRES_PASSWORD)
-  existingSecret: ""
-  # -- External database SSL mode (require, verify-ca, verify-full, disable). Defaults to require for encrypted connections.
-  sslMode: "require"
+  postgres:
+    # -- External PostgreSQL connection string. Used if set, regardless of bundled.enabled.
```
timflannagan: A bit confused on the precedence logic based on this comment and had to look at the templates to figure out the behavior. From my first read: "Used if set, regardless of bundled.enabled" - used for what? The app connection? Does the bundled pod still deploy? Does the init container still run? etc.

charts/agentregistry/values.yaml
```diff
@@ -263,19 +267,36 @@ rbac:
...
+    url: ""
+    # -- Path to a file containing the database URL. Takes precedence over url.
+    urlFile: ""
```
timflannagan: Path to a file mounted on the pod containing the PostgreSQL connection string (e.g. a mounted Secret)? I need to double check the templates, but I don't think setting this field would work today. We don't support setting extra volume mounts on the server deployment atm.

scripts/kind/README.md
```diff
@@ -20,21 +20,20 @@ This single command sets up the full local environment.
...
-2. **`make install-postgresql`** — creates the `agentregistry` namespace and deploys standalone PostgreSQL/pgvector using `pgvector/pgvector:0.8.2-pg16`
-3. **`make install-agentregistry`** — builds server images, pushes them to the local registry, and Helm installs AgentRegistry connected to the local PostgreSQL
+2. **`make install-agentregistry`** — builds server images, pushes them to the local registry, and Helm installs AgentRegistry with a bundled PostgreSQL/pgvector instance
 
 You can also run any sub-target individually, e.g. `make install-agentregistry` to redeploy after a code change.
 
 ## Database Details
 
-The local PostgreSQL instance is configured as follows:
+PostgreSQL/pgvector is bundled in the Helm chart and deployed automatically. The default configuration is:
```
timflannagan: > PostgreSQL/pgvector is bundled in the Helm chart and deployed automatically

Accurate in the context of this scripts' README, but a bit misleading for the default values.yaml?

README.md
```diff
@@ -108,31 +108,30 @@ Open `http://localhost:12121` to use the web UI.
...
   --create-namespace \
-  --set database.host=postgres-pgvector.agentregistry.svc.cluster.local \
-  --set database.password=agentregistry \
-  --set database.sslMode=disable \
+  --set database.postgres.bundled.enabled=false \
+  --set database.postgres.url=postgres://<user>:<password>@<host>:5432/<dbname> \
   --set config.jwtPrivateKey=$(openssl rand -hex 32)
 ```
 
+> **Semantic search** requires a vector-enabled PostgreSQL instance. Add `--set database.postgres.vectorEnabled=true` when your database has vector support. The bundled database does not include vector support — semantic search will not be available when using it.
```
timflannagan: This is buried within the Kubernetes section, but this commentary is also applicable to the registry hosted in docker flow as well.

timflannagan: Should we align the chart & internal/daemon/docker-compose.yml defaults here then?

charts/agentregistry/templates/postgresql-secret.yaml
timflannagan, line 15: The local compose file uses "agent-registry" as well. Not a big deal, just an FYI around inconsistency.

charts/agentregistry/values.yaml
```diff
@@ -263,19 +267,36 @@ rbac:
...
-  database: "agent-registry"
-  # -- External database user
-  username: "agentregistry"
-  # -- External database password (ignored when existingSecret is set)
-  password: ""
-  # -- Name of an existing Secret containing the database password (key: POSTGRES_PASSWORD)
-  existingSecret: ""
-  # -- External database SSL mode (require, verify-ca, verify-full, disable). Defaults to require for encrypted connections.
-  sslMode: "require"
+  postgres:
```
timflannagan, line 278: I'd still advocate for a mode-based pattern here:

```yaml
database:
  mode: bundled
  external: { ... }
  bundled: { ... }
```

Fine for now; this was already discussed offline.

charts/agentregistry/values.yaml
timflannagan: I think we can also remove the AGENT_* envvar references here. Leaks implementation details.

internal/registry/config/config.go
```diff
@@ -82,6 +86,16 @@ func NewConfig() *Config {
 		os.Exit(1)
 	}
 
+	// If a database URL file is specified, read it and override DatabaseURL.
+	if cfg.DatabaseURLFile != "" {
+		url, err := readDatabaseURLFile(cfg.DatabaseURLFile)
+		if err != nil {
+			slog.Error("failed to read database URL file", "path", cfg.DatabaseURLFile, "error", err)
+			os.Exit(1)
```
timflannagan: not blocking: Looks like this is consistent with the rest of this file, but os.Exit(1) makes this implementation hard to test. Existing problem.

charts/agentregistry/values.yaml
nikolasmatt: > This field living under the postgres section is a bit confusing to me. It's effectively configuring fields that live on the server (vs. configuring the postgres instance) and tells it to run vector migrations on startup + enable semantic search functionality. WYDT?

the url field doesn't configure the database either, it configures the registry. The discussion we had offline was to think about this as informational on what the DB is capable and then turn the features that depend on the capability on or off accordingly instead of controlling individual features.

>I think we can also remove the AGENT_* envvar references here. Leaks implementation details.

👍

charts/agentregistry/values.yaml
nikolasmatt: I would push that to a future improvement. We need to figure out how configurable this bundled DB _needs_ to be. For now less is better.

charts/agentregistry/values.yaml
nikolasmatt: I'll update the comment for clarity 👍

charts/agentregistry/values.yaml
nikolasmatt: Removing it for now, will address in followup

charts/agentregistry/templates/postgresql-secret.yaml
nikolasmatt: i updated it all to match

internal/registry/config/config.go
nikolasmatt: this code is removed now so no longer relevant

PR #407, opened 2026-03-24
Makefile: Move Go tools into tools module

### LINKED ISSUE #377: Broken lint Makefile target
Follow-up to #253 where the lint Makefile target is now broken:

```bash
$ make lint
go tool golangci-lint run --fix
go: no such tool "golangci-lint"
make: *** [Makefile:439: lint] Error 2
```

We should fix this and determine the right path forward w.r.t. managing tools going forward since there's inconsistency.

timflannagan:
# Description

We moved executable Go tooling into a dedicated tools module and wired make targets and scripts to run tools through go tool with a modfile override.

The lint workflow now runs our make targets directly and no longer depends on the third-party golangci action. This also committed formatter changes from golangci-lint --fix so the tree matches the new lint behavior.

Fixes #377 

# Change Type

/kind cleanup

# Changelog

```release-note
NONE
```

# Additional Notes


scripts/kind/setup-kind.sh
```diff
@@ -4,9 +4,13 @@ set -o errexit
...
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "${SCRIPT_DIR}/../.." && pwd)"
 
-# Use kind via go tool directives (go.mod) so no separate install step is needed.
-KIND=(go tool kind)
+# Use kind via go tool directives (tools/go.mod) so no separate install step is needed.
+# Override KIND to use a different binary or command.
+TOOLS_MODFILE="${TOOLS_MODFILE:-${REPO_ROOT}/tools/go.mod}"
+KIND="${KIND:-go tool -modfile=${TOOLS_MODFILE} kind}"
+read -r -a KIND_TOOL <<<"${KIND}"
```
timflannagan, line 13: TODO: AI generated, but this should be improved.

tools/go.mod
```diff
@@ -0,0 +1,229 @@
+module github.com/agentregistry-dev/agentregistry/tools
+
+go 1.25.7
+
+tool (
```
timflannagan, line 5: We should add helm as well in the future

.github/workflows/lint.yml
```diff
@@ -19,9 +19,8 @@ jobs:
         with:
           go-version-file: "go.mod"
 
-      - uses: golangci/golangci-lint-action@v7
-        with:
-          version: v2.8.0
+      - name: Lint Go
+        run: make lint
```
timflannagan, line 23: TODO: Override the arg variable since --fix is enabled by default

Issue #404, opened 2026-03-20
Failed to create MCP server


wigging:
I'm using the following version of the agent registry:

```
arctl version v0.3.2
Git commit: 43df00d
Build date: 2026-03-16
Server version: v0.3.2
Server git commit: 43df00d
Server build date: 2026-03-16
```

I tried to add an MCP server in the UI by pointing it to a GitLab repository but it says "failed to create server". The GitLab repo contains a MCP server that is built with the Python FastMCP framework. How can I add an MCP server to the registry?

<img width="1103" height="751" alt="Image" src="https://github.com/user-attachments/assets/7684481d-af6e-42b0-80b4-e635bda6c822" />

wigging:
@peterj This might be related to issue #402 which you also submitted last week.

lleon-at-navteca:
I reviewed this together with PR #408 and issue #405.

My conclusion is that PR #408 is very likely related to the failure reported here, but I would describe it as a partial or indirect fix rather than a guaranteed full resolution of #404.

Why I say that:
- this issue reports the symptom: "Failed to create MCP server"
- PR #408 fixes one concrete cause of that symptom in the Add Server UI:
  - missing required transport URLs for `sse` / `streamable-http`
  - incorrect package defaulting (`npm` vs `oci`)

So if the server creation failed because the form generated an invalid transport/package payload, then PR #408 would address that.

However, this screenshot alone does not prove that this exact invalid-payload path was the only cause in this issue. The issue is broader in wording than the specific bug fixed in #408.

I would say:
- #408 definitely fixes #405
- #408 may also have resolved some cases that surfaced as #404
- but I would be cautious about closing #404 based only on #408 unless the original reproduction is re-tested against that branch

Issue #402, opened 2026-03-20
can't deploy remote MCP server from the UI


peterj:
the deploy button is disabled for remote mcp servers (this is probably a simple fix we need in the UI)

Abhishek9639:
Hi @peterj, 
I’d like to work on this issue. I can look into why the deploy button is disabled for remote MCP servers and fix it in the UI. 
Could you please assign this to me?
Thanks

lleon-at-navteca:
I reviewed this issue together with PR #408 and the current UI behavior.

My view is that PR #408 does not directly fix #402 as written, but it may incidentally unblock some cases.

Reason:
- this issue is about the Deploy button being disabled for remote MCP servers
- the Deploy button logic is tied to whether the server has an OCI package available
- PR #408 changes the Add Server form so that package entries default to `oci` and can be submitted correctly when a required transport URL is present

Because of that, PR #408 can help in a narrower scenario:
- if the server was being registered incorrectly before, and therefore ended up without a valid OCI package, then after #408 the same server may be created in a deployable form and the button may appear enabled

But that is different from saying the original Deploy-button logic for remote servers was fixed directly.

So I would describe #408 as related, and possibly unblocking some reproductions, but not as a complete direct fix for #402 itself.

PR #408, opened 2026-03-25
Enhance Add Server dialog with transport URL validation and OCI support

### LINKED ISSUE #405: UI is allowing creating invalid payloads for MCP servers using streamable-http or sse transports
The Add Server UI allowed creating invalid payloads for MCP servers using streamable-http or sse transports, which caused backend validation errors on publish. It also defaulted package registryType to npm, which does not match the containerized MCP server flow.

lleon-at-navteca:
# Description

Fixes the Add Server dialog so package transport validation behaves correctly for URL-based transports and improves default package configuration.

Resolves #405  

**What changed:**
- **Transport URL validation** — Prevents submission when `sse` or `streamable-http` is selected without a required URL.
- **OCI as default registry** — New packages now default to the `oci` registry type.
- **Improved package transport UI** — Package transport selection now uses radio buttons for `stdio`, `sse`, and `streamable-http`.
- **Conditional transport URL input** — The dialog now shows a transport URL field for package transports that require it.
- **Improved package registry selection** — Package registry type selection includes `oci`, `npm`, `pypi`, and `docker`, with `oci` selected by default.
- **Canonical schema URL** — Updated the default schema value to the full MCP server schema URL.
- **Unit tests** — Added coverage for package defaults, transport URL validation, and payload generation for `streamable-http`.

# Change Type

/kind fix

# Changelog

```release-note
Fixed Add Server dialog validation for URL-based package transports and improved default package settings.
```

# Additional Notes

- Verified the new AddServerDialog test file passes in isolation.
- Verified the full UI test suite passes.
- The new tests cover both validation failures and successful payload generation for `streamable-http`.
package-form

### DIFF
--- a/ui/components/__tests__/add-server-dialog.test.tsx
+++ b/ui/components/__tests__/add-server-dialog.test.tsx
@@ -1,0 +1,121 @@
+import { render, screen, waitFor } from "@testing-library/react"
+import userEvent from "@testing-library/user-event"
+import { beforeEach, describe, expect, it, vi } from "vitest"
+
+import { AddServerDialog } from "../add-server-dialog"
+import { createServerV0 } from "@/lib/admin-api"
+import { toast } from "sonner"
+
+vi.mock("@/lib/admin-api", () => ({
+  createServerV0: vi.fn(),
+}))
+
+vi.mock("sonner", () => ({
+  toast: {
+    success: vi.fn(),
+    error: vi.fn(),
+  },
+}))
+
+describe("AddServerDialog", () => {
+  beforeEach(() => {
+    vi.clearAllMocks()
+    vi.mocked(createServerV0).mockResolvedValue({
+      data: { server: { name: "io.navteca/hello-mcp" } },
+    } as never)
+  })
+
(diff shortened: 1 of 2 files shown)


lleon-at-navteca:
I revisited the full scenario around this PR and the related issues.

This PR clearly fixes the problem described in #405.

Why:
- Before this change, the Add Server dialog allowed users to submit `sse` and `streamable-http` transports without the required URL.
- It also defaulted package entries to `npm`, which does not fit the OCI/container flow used for deployable MCP servers.
- This PR fixes both sides in the UI by:
  - requiring a transport URL for `sse` and `streamable-http`
  - including that URL in the generated payload
  - defaulting package `registryType` to `oci`

That means the UI now prevents the invalid payloads that were causing backend validation failures.

This PR is also related to #404 and #402, but more indirectly:
- it may unblock some cases behind those reports by allowing users to register a valid OCI-backed server definition
- however, I would not claim it fully resolves those issues as written, because they describe broader symptoms than the specific invalid-payload bug fixed here

PR #413, opened 2026-04-01
[Proposal] Use annotation-based daemon skip for cobra commands

inFocus7:
This is a proposal, can be removed if not as helpful as I think it is 👍🏼

# Description

- Remove daemon skip map
- Use special annotation to skip daemon on commands

Currently we define the set of skipped commands in an un-exported field. If anyone extends this library they won't be able to easily set up commands that don't require a daemon running. We can create extension points (exported functionality or cli option one can plug in commands), but this would be _another_ extension point which adds complexity to the codebase.

Proposal is to move from the map-based skips to an annotation-based one. Commands would be able to define if they don't need the daemon directly in its definition through the annotation field.

> Annotations are key/value pairs that can be used by applications to identify or group commands or set special options.

# Change Type

```
/kind cleanup
```

# Changelog

```release-note
NONE
```

# Additional Notes

An edge case would be if commands define their own `PersistentPreRun`, then the one we use now is overriden and so setting `skipDaemon: true` on those commands won't have any effect. This is a pre-existing edge case, but worth noting.

The root `completion` command is from cobra, so it is the exception where we can't directly define the annotation for.

### DIFF
--- a/pkg/cli/annotations/annotations.go
+++ b/pkg/cli/annotations/annotations.go
@@ -1,0 +1,8 @@
+package annotations
+
+// SkipDaemonAnnotation is the cobra.Command annotation key that opts a command
+// out of PersistentPreRunE setup (daemon connection, API client creation, etc.).
+// Add it to any command's Annotations map with the value "true" to skip.
+//
+//	Annotations: map[string]string{annotations.SkipDaemonAnnotation: "true"}
+const SkipDaemonAnnotation = "skipDaemon"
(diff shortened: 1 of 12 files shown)


pkg/cli/annotations/annotations.go
```diff
@@ -0,0 +1,8 @@
+package annotations
+
+// SkipDaemonAnnotation is the cobra.Command annotation key that opts a command
+// out of PersistentPreRunE setup (daemon connection, API client creation, etc.).
+// Add it to any command's Annotations map with the value "true" to skip.
+//
+//	Annotations: map[string]string{annotations.SkipDaemonAnnotation: "true"}
+const SkipDaemonAnnotation = "skipDaemon"
```
peterj, line 8: nit: should this be just called "SkipDaemon" ? since it's in the annotation package already

peterj:
I like it - much cleaner compared to what we have.

PR #425, opened 2026-04-14
feat: add declarative CLI with server-side apply for all resource types

xytian315:
# Description

 **Motivation**

Users needed a kubectl-style declarative workflow to manage Agent Registry resources — defining agents, MCP servers, skills, prompts, and providers as YAML files and applying them with a single command, rather than calling imperative API commands directly. `arctl apply -f agent.yaml` should be idempotent: running it twice produces the same result with no errors.

**What changed**

**Declarative CLI commands:**
  - `arctl apply -f <file>` — applies one or more YAML documents (multi-doc files supported); 
  - `arctl get <resource> [name] `— lists or fetches a single resource; supports -o yaml for full output
  - `arctl delete -f <file>` and `arctl delete <kind> <name> --version` — removes resources declaratively or explicitly
  - `arctl init agent|mcp|skill|prompt `— scaffolds a project directory and generates a typed YAML manifest
  - `arctl build <dir> --push` — builds and pushes a Docker image for agent/MCP/skill projects

 **Server-side apply (PUT endpoints):**
  - PUT /v0/agents/{name}/versions/{version} — idempotent upsert for agents
  - PUT /v0/skills/{name}/versions/{version} — idempotent upsert for skills
  - PUT /v0/prompts/{name}/versions/{version} — idempotent upsert for prompts
  - PUT /v0/servers/{name}/versions/{version} — idempotent upsert for MCP servers (admin edit moved from PUT to PATCH)
  - PUT /v0/providers/{id}?platform=xxx — idempotent upsert for providers (creates when platform provided + provider missing; updates
  otherwise)
  - PUT /v0/agents/{name}/versions/{version}/deployments/{providerId} — idempotent agent deployment (no-op if already deployed with same
  env/config; undeploy + relaunch on drift)
  - PUT /v0/servers/{name}/versions/{version}/deployments/{providerId} — idempotent server deployment

**Service layer (follows phase-1 service/db refactor):**
  - Apply* methods for all 6 resource types (agent, skill, prompt, server, provider, deployment) — check exists → update or create,
  idempotent
  - Deployment apply detects env/providerConfig drift and redeploys when changed; no-op when identical
  - UpdatePrompt added to PromptStore (previously had no update method)

**Others**
  - Unit, integration, and e2e test coverage for all apply paths
  - docs/declarative-cli.md and examples/declarative/ with working example YAML files for all resource types
  - openapi.yaml updated with all new PUT endpoints
  - Fixed Mac Docker buildx pushes by routing them through a dedicated buildnet network to kind-registry:5000
  
# Change Type

/kind feature

# Changelog

```release-note
feat: add declarative CLI with server-side apply for all resource types
```

# Additional Notes

Refer to `docs/declarative-cli.md` for manual CLI testing.
```
 ## Agents
  arctl init agent adk python summarizer --model-provider gemini --model-name gemini-2.0-flash
  arctl build summarizer/ --push
  arctl apply -f summarizer/agent.yaml
  arctl apply -f summarizer/agent.yaml   # second apply: idempotent, no error
  arctl get agent summarizer
  arctl delete agent summarizer --version 0.1.0

  ## MCP Servers
  arctl init mcp fastmcp-python acme/my-server
  arctl build my-server/ --push
  arctl apply -f my-server/mcp.yaml
  arctl apply -f my-server/mcp.yaml      # idempotent
  arctl get mcps
  arctl delete mcp acme/my-server --version 0.1.0

  ## Skills & Prompts
  arctl init skill summarize --category nlp
  arctl apply -f summarize/skill.yaml
  arctl apply -f summarize/skill.yaml    # idempotent
  arctl get skills

  arctl init prompt summarizer-system-prompt
  arctl apply -f summarizer-system-prompt.yaml
  arctl get prompts

  ## Apply multiple resources from one file
  arctl apply -f examples/declarative/full-stack.yaml
  arctl apply -f examples/declarative/full-stack.yaml  # fully idempotent
  arctl get all
  arctl delete -f examples/declarative/full-stack.yaml
```
  **Known limitation (follow-up)**: `arctl agent run .` does not work with declarative agent.yaml — it uses the old flat format. Needs a separate fix.


internal/registry/api/handlers/v0/agents/deployment_handlers.go
```diff
@@ -0,0 +1,86 @@
...
+		if err != nil {
+			return nil, huma.Error400BadRequest("Invalid version encoding", err)
+		}
+		providerID := strings.TrimSpace(input.ProviderID)
+		if providerID == "" {
+			return nil, huma.Error400BadRequest("providerId is required")
+		}
+
+		deployment, err := deploymentSvc.ApplyAgentDeployment(ctx, agentName, version, providerID, input.Body.Env, input.Body.ProviderConfig)
+		if err != nil {
```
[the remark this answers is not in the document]
xytian315: fixed.

internal/registry/api/handlers/v0/servers/deployment_handlers.go
```diff
@@ -0,0 +1,86 @@
...
+		if err != nil {
+			return nil, huma.Error400BadRequest("Invalid version encoding", err)
+		}
+		providerID := strings.TrimSpace(input.ProviderID)
+		if providerID == "" {
+			return nil, huma.Error400BadRequest("providerId is required")
+		}
+
+		deployment, err := deploymentSvc.ApplyServerDeployment(ctx, serverName, version, providerID, input.Body.Env, input.Body.ProviderConfig)
+		if err != nil {
```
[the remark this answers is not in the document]
xytian315: fixed

docs/declarative-cli.md
```diff
@@ -0,0 +1,67 @@
+# Declarative CLI
+
+Define agents, MCP servers, skills, and prompts as YAML files and manage them with `arctl apply`, `arctl get`, and `arctl delete`.
```
shashankram, line 3: We support STDIN right? `-f - <<EOF`

internal/registry/api/handlers/v0/servers/deployment_handlers.go
```diff
@@ -0,0 +1,86 @@
+package servers
```
shashankram, line 1: Can we rename servers to mcpservers everywhere?

shashankram: Not a problem if it's a hassle and a big change. Just don't like using a generic term like `servers` for this

docs/declarative-cli.md
xytian315: yes. I can call it out in the doc.

internal/registry/api/handlers/v0/servers/deployment_handlers.go
xytian315: I totally agree the name is confusing!! Could we please create an issue and do in a follow up PR? 

I would like to keep this PR not touching the existing API/Service/DB etc as much as possible.
 
And I think we would want to change service/db/router naming as well too e.g. internal/registry/service/server/ so the scope to change this is a bit big. If you agree, I can create an issue to track this. 🙇‍♀️

shashankram: Makes sense. We can revisit this later

internal/registry/api/handlers/v0/servers/edit.go
```diff
@@ -30,10 +30,10 @@ func RegisterEditEndpoints(api huma.API, pathPrefix string, serverSvc serversvc.
 	// Edit server endpoint
 	huma.Register(api, huma.Operation{
 		OperationID: "edit-server" + strings.ReplaceAll(pathPrefix, "/", "-"),
-		Method:      http.MethodPut,
+		Method:      http.MethodPatch,
```
xytian315, line 33: Calling out the only change to the existing API. I changed this from PUT to PATCH so I can use PUT for apply following the same pattern as other apply resources design. 

The upstream MCP registry (github.com/modelcontextprotocol/registry) originally defined PUT /v0/servers/{name}/versions/{version} as the admin edit endpoint, When we forked that code into this repo, we inherited RegisterEditEndpoints at internal/registry/api/handlers/v0/servers/edit.go,  We couldn't have both on the same method+path (Huma rejects duplicate registrations), so we moved the inherited admin edit to PATCH and gave PUT to apply. 

Please let me know if you have any concern.

examples/declarative/full-stack.yaml
```diff
@@ -0,0 +1,74 @@
...
+  description: "Default system prompt for summarization agents"
+  content: |
+    You are a document summarization assistant.
+    Keep summaries concise and accurate.
+---
+apiVersion: ar.dev/v1alpha1
+kind: Agent
+metadata:
+  name: summarizer
+  version: "1.0.0"
```
shashankram, line 51: Where is this k8s type structure coming from? Are these accurate?

xytian315: from `internal/cli/scheme/scheme.go` https://github.com/xytian315/agentregistry/blob/ec7119116db239030336c01ab22da0212b02ead2/internal/cli/scheme/scheme.go

The spec fields matched `pkg/models/manifest.go` https://github.com/xytian315/agentregistry/blob/ec7119116db239030336c01ab22da0212b02ead2/pkg/models/manifest.go

internal/cli/scheme/scheme.go
```diff
@@ -0,0 +1,83 @@
...
+	APIVersion string         `yaml:"apiVersion"`
+	Kind       string         `yaml:"kind"`
+	Metadata   Metadata       `yaml:"metadata"`
+	Spec       map[string]any `yaml:"spec"`
+}
+
+// Metadata holds the resource identity and server-generated timestamps.
+type Metadata struct {
+	Name        string     `yaml:"name"`
+	Version     string     `yaml:"version,omitempty"`
```
shashankram, line 26: Why do we need a Version if the SSA behavior is to upsert?

xytian315: The resource identity is (name, version) as defined by the [DB primary key](https://github.com/xytian315/agentregistry/blob/feat-add-declarative-cli-refactor/internal/registry/database/migrations/001_initial_schema.sql#L38)

Resources like agent, mcpserver and skills all identified by name and version. provider resource is identified by ID only, no version.

e2e/e2e_test.go
```diff
@@ -78,7 +90,67 @@ func discoverRegistryURL() string {
 		log.Fatalf("LoadBalancer IP not assigned to agentregistry service within %s — is MetalLB running?", timeout)
 	}
 
-	return fmt.Sprintf("http://%s:12121/v0", ip)
+	if runtime.GOOS == "darwin" {
+		log.Printf("macOS: LoadBalancer IP %s not routable from host — using port-forward", ip)
+		return startPortForward(servicePort)
```
shashankram, line 95: Who stops the PortForward

shashankram: Can a Context be passed to this that is canceled before the test ends?

shashankram (approved):

e2e/e2e_test.go
xytian315: startPortForward func returns a stop func https://github.com/xytian315/agentregistry/blob/ec7119116db239030336c01ab22da0212b02ead2/e2e/e2e_test.go#L153

do you prefer me to refactor this func to pass the context in?

internal/registry/api/handlers/v0/agents/deployment_handlers.go
```diff
@@ -0,0 +1,86 @@
...
+		version, err := url.PathUnescape(input.Version)
+		if err != nil {
+			return nil, huma.Error400BadRequest("Invalid version encoding", err)
+		}
+		providerID := strings.TrimSpace(input.ProviderID)
+		if providerID == "" {
+			return nil, huma.Error400BadRequest("providerId is required")
+		}
+
+		deployment, err := deploymentSvc.ApplyAgentDeployment(ctx, agentName, version, providerID, input.Body.Env, input.Body.ProviderConfig)
```
ilackarms, line 61: **Blocking:** `PreferRemote` is accepted in `DeploymentApplyBody` but never passed to the service layer — `ApplyAgentDeployment` and `ApplyServerDeployment` in `records.go:309-314` hardcode it to `false`. The API documents a field that has no effect.

Fix: either thread `preferRemote` through the service method signature, or remove it from `DeploymentApplyBody`.

internal/registry/service/deployment/records.go
```diff
@@ -268,3 +270,76 @@ func (s *registry) removeDeploymentRecord(ctx context.Context, deployment *model
...
+		return nil, fmt.Errorf("checking existing deployment: %w", err)
+	}
+	if existing != nil && existing.Status == models.DeploymentStatusDeployed &&
+		envEqual(existing.Env, env) &&
+		providerConfigEqual(existing.ProviderConfig, providerConfig) {
+		return existing, nil // identical desired state — idempotent no-op
+	}
+
+	// Clean up stale record (failed/cancelled/deploying/drift) before launching fresh.
+	if err := s.CleanupExistingDeployment(ctx, resourceName, version, resourceType, providerID); err != nil {
```
ilackarms, line 294: **Blocking:** This code silently undeploys live deployments when config drifts:

1. **Footgun:** A typo in `env` silently tears down a running deployment. Should require `force` flag — without it, error with "deployment exists with different config; use --force to replace".
2. **Silent failure:** If platform undeploy fails, cleanup only logs a warning (line 71-72) and still deletes the DB record. `LaunchDeployment` then creates a second deployment while the first is still running.
3. **Race condition:** No transaction wraps cleanup + launch (lines 294-306). Concurrent apply can slip between delete and create.
4. **Redundant lookup:** When `existing` is nil, cleanup does a second `findDeploymentByIdentity` that also finds nothing. Gate on `existing != nil`.

Recommend: default to erroring on drift, require `force` flag to proceed with undeploy + relaunch.

internal/registry/api/router/v0.go
```diff
@@ -71,16 +71,22 @@ func RegisterRoutes(
 	v0version.RegisterVersionEndpoint(api, pathPrefix, versionInfo)
 	v0servers.RegisterServersEndpoints(api, pathPrefix, svcs.Server, svcs.Deployment)
 	v0servers.RegisterServersCreateEndpoint(api, pathPrefix, svcs.Server, svcs.Deployment)
+	v0servers.RegisterServersApplyEndpoint(api, pathPrefix, svcs.Server, svcs.Deployment)
```
ilackarms, line 74: **Design:** The intent for server-side apply was a single batch endpoint — `POST /v0/apply` — that accepts a multi-resource YAML/JSON document. The server parses by kind and dispatches to the internal `Apply*` service methods directly.

The current design pushes orchestration to the client: the CLI loops and makes N individual HTTP calls to per-kind PUT endpoints. Every client must know every kind's URL pattern, and adding a new kind requires a new endpoint + handler + client method.

A single batch endpoint: one endpoint to maintain, clients just POST documents, new kinds only need a service method + registration, enables transactional batching.

The per-resource PUT endpoints aren't wrong as a lower-level API, but they shouldn't be the primary apply surface.

internal/cli/scheme/scheme.go
```diff
@@ -0,0 +1,83 @@
...
+)
+
+const APIVersion = "ar.dev/v1alpha1"
+
+// Resource is the universal declarative envelope for all registry resources.
+type Resource struct {
+	APIVersion string         `yaml:"apiVersion"`
+	Kind       string         `yaml:"kind"`
+	Metadata   Metadata       `yaml:"metadata"`
+	Spec       map[string]any `yaml:"spec"`
```
ilackarms, line 20: **Design:** `Spec map[string]any` flows through the entire handler layer with no compile-time enforcement. The JSON roundtrip in each `to*JSON` method is the only bridge to typed code.

Our convention: API object fields should use concrete typed structs so the contract is enforced at compile time. See `platforms/kubernetes/deployment_adapter_kubernetes_platform.go:64-76` for the `UnmarshalInto` pattern.

internal/cli/resource/handler.go
```diff
@@ -0,0 +1,115 @@
...
+	"slices"
+	"strings"
+	"sync"
+
+	"github.com/agentregistry-dev/agentregistry/internal/cli/scheme"
+	"github.com/agentregistry-dev/agentregistry/internal/client"
+)
+
+// ResourceHandler translates between declarative YAML Resources and the registry API.
+type ResourceHandler interface {
```
ilackarms, line 14: **Design:** The four implementations (`agent.go`, `mcpserver.go`, `skill.go`, `prompt.go`) are nearly identical boilerplate — same JSON roundtrip, same `ToResource` field-stripping, same table formatting. Only differences: kind string, typed struct, client method, column layout.

Should be a single generic `ResourceHandler` struct driven by schema definitions. Adding a new kind should be a registration call, not a new file — similar to how K8s CRD controllers work off GVK + schema.

internal/registry/api/handlers/v0/agents/handlers.go
```diff
@@ -279,3 +279,51 @@ func RegisterAgentsCreateEndpoint(api huma.API, pathPrefix string, agentSvc agen
...
 }
+
+// ApplyAgentInput represents the input for applying (create or update) a specific agent version
+type ApplyAgentInput struct {
+	AgentName string                `path:"agentName"`
+	Version   string                `path:"version"`
+	Body      agentmodels.AgentJSON `body:""`
+}
+
+func RegisterAgentsApplyEndpoint(api huma.API, pathPrefix string, agentSvc agentsvc.Registry, deploymentSvc deploymentmeta.Lister) {
```
ilackarms, line 290: **Design:** This apply handler is structurally identical to the ones in `servers/handlers.go`, `skills/handlers.go`, and `prompts/handlers.go` — define `Apply*Input`, register PUT, unescape path params, set `Body.Name`/`Body.Version`, call `svc.Apply*()`, map errors.

Same per-kind boilerplate concern as the resource handlers (https://github.com/agentregistry-dev/agentregistry/pull/425#discussion_r3083687305). A single batch endpoint (https://github.com/agentregistry-dev/agentregistry/pull/425#discussion_r3083686340) would eliminate all four of these.

internal/cli/declarative/init.go
```diff
@@ -0,0 +1,638 @@
...
+		return registryName[i+1:]
+	}
+	return registryName
+}
+
+// writeDeclarativeAgentYAML writes agent.yaml in the ar.dev/v1alpha1 declarative format.
+func writeDeclarativeAgentYAML(projectDir, name, ver, image, language, framework, modelProvider, modelName, description, gitURL string, mcps, skills, prompts []string) error {
+	registryURL := agentutils.GetDefaultRegistryURL()
+
+	spec := map[string]any{
```
ilackarms, line 199: **Design:** Same `map[string]any` concern as `scheme.go:20` (https://github.com/agentregistry-dev/agentregistry/pull/425#discussion_r3083686753), but on the write side. Builds spec with hardcoded string keys (`"image"`, `"language"`, `"framework"`, etc.). If a field name changes in `AgentJSON`, these literals silently produce invalid YAML.

Should populate a typed struct and marshal that, so renames are caught at compile time. Same applies to `writeDeclarativeMCPYAML`, `writeDeclarativeSkillYAML`, and `writeDeclarativePromptYAML`.

internal/cli/declarative/build.go
```diff
@@ -0,0 +1,250 @@
...
+				return fmt.Errorf("expected a project directory, not a file — try: arctl build ./my-project")
+			}
+
+			r, yamlFile, err := findDeclarativeResource(projectDir)
+			if err != nil {
+				return err
+			}
+
+			out := cmd.OutOrStdout()
+			switch r.Kind {
```
ilackarms, line 68: **Design (minor):** Another per-kind fork maintained in sync with the resource registry. Build implementations are genuinely kind-specific, but dispatch and image resolution (`resolveImage`, `specString`, `specPackageIdentifier`) could be driven by a build config registered per kind rather than a switch.

internal/registry/api/handlers/v0/agents/deployment_handlers.go
```diff
@@ -0,0 +1,86 @@
...
+		if err != nil {
+			return nil, applyDeploymentHTTPError(err)
+		}
+		return &DeploymentApplyResponse{Body: *deployment}, nil
+	})
+}
+
+func applyDeploymentHTTPError(err error) error {
```
ilackarms, line 69: **Nit:** `applyDeploymentHTTPError` is identical to `serverApplyDeploymentHTTPError` in `servers/deployment_handlers.go`. Extract to a shared helper.

internal/registry/service/deployment/records.go
```diff
@@ -268,3 +270,76 @@ func (s *registry) removeDeploymentRecord(ctx context.Context, deployment *model
...
+		return nil, fmt.Errorf("%w: resource name, version, and provider ID are required", database.ErrInvalidInput)
+	}
+
+	existing, err := s.findDeploymentByIdentity(ctx, resourceName, version, resourceType, providerID)
+	if err != nil && !errors.Is(err, database.ErrNotFound) {
+		return nil, fmt.Errorf("checking existing deployment: %w", err)
+	}
+	if existing != nil && existing.Status == models.DeploymentStatusDeployed &&
+		envEqual(existing.Env, env) &&
+		providerConfigEqual(existing.ProviderConfig, providerConfig) {
```
ilackarms, line 289: **Suggestion:** Drift detection compares `env` and `providerConfig` but not `preferRemote`. Changing only `preferRemote` won't trigger redeployment.

PR #434, opened 2026-04-16
refactor(apply): replace per-kind PUT endpoints with POST /v0/apply batch endpoint

xytian315:
# Description
refactor(apply): replace per-kind PUT endpoints with POST /v0/apply batch endpoint

Adressed all the comments from this PR #425 

  **Motivation:** PR #425 landed declarative CLI with per-kind PUT endpoints. @ilackarms review identified architectural improvements: a batch endpoint as the primary apply surface to eliminate per-kind boilerplates

# Change Type

/kind cleanup
# Changelog

```release-note
refactor(apply): replace per-kind PUT endpoints with POST /v0/apply batch endpoint
```

# Additional Notes
- All declarative CLI commands (apply, delete, get, init, build) work as before. refer to [doc/declarative-cli.md ](https://github.com/agentregistry-dev/agentregistry/blob/main/docs/declarative-cli.md) for testing.

PR #436, opened 2026-04-16
mcp: dynamic mcp resolution via registry refs and env var injection for MCP config

shashankram:
# Description
- Enables injecting the MCP config via MCP_SERVERS_CONFIG env var to the agent.
- Allows ref based resolution of MCP servers using just name+version instead of inlining the entire spec.
- Allows bypassing the MCP hostname check using `INSECURE_SKIP_MCP_HOST_VERIFY` env var to enabe
  testing MCP servers on localhost.

# Change Type

```
/kind cleanup
```

# Changelog

```release-note
None
```


internal/constants/constants.go
```diff
@@ -0,0 +1,29 @@
+// Package constants defines shared constant values used across the agentregistry codebase.
+package constants
+
+// Agent runtime environment variable keys injected by ResolveAgent and consumed
+// by platform adapters and agent frameworks at deploy/run time.
+const (
+	// EnvKagentNamespace is the Kubernetes namespace where the agent is deployed.
+	// Defaults to "default" when not explicitly set.
+	EnvKagentNamespace = "KAGENT_NAMESPACE"
```
timflannagan, line 9: nit: Seeing some tests still hardcode this variable in /work/agentregistry/internal/registry/platforms/kubernetes/deployment_adapter_kubernetes_platform_test.go & /work/agentregistry/internal/registry/platforms/utils/deployment_adapter_utils_test.go.

internal/registry/validators/utils.go
```diff
@@ -132,10 +149,13 @@ func IsValidRemoteURL(rawURL string) bool {
 		return false
 	}
 
-	// Reject localhost URLs for remotes (security/production concerns)
-	hostname := u.Hostname()
-	if hostname == "localhost" || hostname == "127.0.0.1" || strings.HasSuffix(hostname, ".localhost") {
-		return false
+	// Reject localhost URLs for remotes (security/production concerns).
+	// This check is bypassed when INSECURE_SKIP_MCP_HOST_VERIFY=true (dev-only).
+	if !shouldSkipMCPHostVerify() {
```
timflannagan, line 154: Should we log a warning when INSECURE_SKIP_MCP_HOST_VERIFY is enabled?

internal/registry/database/postgres_agents.go
```diff
@@ -365,9 +419,24 @@ func (s *agentStore) CreateAgent(ctx context.Context, agentJSON *models.AgentJSO
 	if err != nil {
 		return nil, fmt.Errorf("failed to marshal agent JSON: %w", err)
 	}
+
+	mcpRefs, skillRefs, promptRefs := extractRegistryRefs(agentJSON)
```
timflannagan, line 423: Add a TODO that skillRefs & promptRefs will be nil until we migrate over to the new UX?

timflannagan (approved):

internal/registry/validators/utils.go
shashankram: It's an undocumented env var for testing, so not required

internal/registry/database/postgres_agents.go
shashankram: Will be handled in https://github.com/agentregistry-dev/agentregistry/pull/434

internal/constants/constants.go
shashankram: Will consolidate later

ilackarms (approved):

PR #440, opened 2026-04-17
Fix resolve for local agent build, it wasn't using auth so failed when an agent had MCP, Prompt, or Skill

### LINKED ISSUE #439: arctl agent run: registry-type MCP server resolution uses unauthenticated client
## Summary

`arctl agent run` fails with `401 Unauthorized` when the agent manifest contains a `type: registry` MCP server. At run time, the CLI must call the registry API to resolve the server spec (image, transport, env vars) before it can start the container. The resolver creates a bare HTTP client that never sends an auth token, so this resolution always fails against any server that requires authentication.

When the manifest has no MCP servers (`mcpServers: []`), the resolver loop is a no-op and the agent starts fine. The bug only surfaces after `arctl agent add-mcp` adds a `type: registry` entry.

## Reproduce

```bash
arctl mcp publish ./my-server --type oci --package-id ... --version 0.1.0
arctl agent add-mcp foo --registry-server-name myorg/foo --registry-server-version 0.1.0
arctl agent run ./my-agent
# → failed to fetch server "myorg/foo" from registry: unexpected status code 401: Unauthorized
```

## Root cause

`internal/cli/agent/utils/registry_resolver.go:92` creates `registry.NewClient()`, which is a plain `http.Client` with no auth:

```go
client := registry.NewClient()  // no token
serverEntry, err := client.FetchServer(registryURL, ...)
```

Every other CLI API call goes through `internal/client.Client` (set via `agent.SetAPIClient()` in `pkg/cli/root.go:77`), which attaches `Authorization: Bearer <token>` on every request (`client.go:104`). The registry resolver bypasses this entirely.

The prompt resolution path at `registry_resolver.go:235` has the same bug — it creates `client.NewClient(registryURL, "")` with an empty token.

## Fix

Pass the authenticated `apiClient` (or its token) into `ParseAgentManifestServers` / `resolveRegistryServer` instead of creating a new unauthenticated client. The prompt resolution path (`ResolveManifestPrompts`) should be fixed the same way.

## Workaround

None — `ARCTL_API_TOKEN` env var and `--registry-token` flag only configure the main `apiClient`, not the `registry.Client` used in the resolution code path.

josh-pritchard:
# Description
Resolve https://github.com/agentregistry-dev/agentregistry/issues/439

# Change Type
/kind fix

# Changelog

```release-note
Fix resolve for local agent build, it wasn't using auth so failed when an agent had MCP, Prompt, or Skill
```

# Additional Notes

### DIFF
--- a/internal/cli/agent/utils/registry_resolver.go
+++ b/internal/cli/agent/utils/registry_resolver.go
@@ -8,12 +8,13 @@ import (
 
 	"github.com/agentregistry-dev/agentregistry/internal/cli/agent/frameworks/common"
 	"github.com/agentregistry-dev/agentregistry/internal/client"
-	"github.com/agentregistry-dev/agentregistry/internal/registry"
 	"github.com/agentregistry-dev/agentregistry/pkg/models"
 	"github.com/modelcontextprotocol/registry/pkg/model"
 )
 
 var defaultRegistryURL = "http://127.0.0.1:12121"
+
+var registryAPIClient *client.Client
 
 // SetDefaultRegistryURL overrides the fallback registry URL used when manifests omit registry_url.
 func SetDefaultRegistryURL(url string) {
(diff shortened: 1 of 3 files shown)


josh-pritchard:
This code is actually slated to be likely cleaned up with later refactor to remove imperative CLI. Merging this will enable some things i am doing but we don't need to be very critical.

PR #449, opened 2026-04-21
Refactor v1alpha1 types

ilackarms:
# Description                                                                                                                                                         
                                                                               
**Motivation:** Legacy registry = one bespoke type/handler/service/table per kind (Agent, MCPServer, Skill, Prompt, Provider, Deployment). Every new kind = N-way     
copy-paste. Wire/storage/service contracts tangled. This PR ports OSS registry in-place to single k8s-style `v1alpha1` envelope (`apiVersion + kind + metadata + spec + status`). Same shape YAML → HTTP → Go client → store → DB.                                                                                                          
                                                                                                                                                                    
**What changed:**                                                                                                                                                     
- **New API pkg** `pkg/api/v1alpha1`: typed objects + `RawObject`, `Scheme` w/ multi-doc decode, `ObjectMeta`
(namespace/name/version/generation/labels/annotations/finalizers/deletionTimestamp), `Status.Conditions`, per-kind `Validate()`/`ResolveRefs()`, shared registry      
validators (OCI/NPM/PyPI/NuGet/MCPB) behind `Dispatcher`.
- **One generic handler** at `/v0/namespaces/{namespace}/{pluralKind}/{name}/{version}` + cross-ns list + multi-doc `POST /v0/apply` + `POST /v0/import`. Kills       
per-kind handlers (agents/skills/prompts/providers/servers/deployments/apply/embeddings/deploymentmeta).                                                              
- **One generic store** (`store_v1alpha1.go`): one JSONB table per kind in `v1alpha1` schema, PK `(namespace,name,version)`, partial-unique `is_latest_version`, GIN
on labels+spec, status-only NOTIFY trigger, soft-delete via `deletionTimestamp`+finalizers. Drops all `postgres_*.go` + `public.*` migrations.                        
- **Deployment** collapsed onto `DeploymentAdapter` iface (async Apply/Remove/Logs/Discover). New coordinator + noop reference adapter. Local + kubernetes adapters
ported, share translate helpers in `platforms/utils`.                                                                                                                 
- **Importer** rebuilt at `pkg/importer`: pipeline (decode → validate → enrich → upsert), `Scanner` plug-in iface, `FindingsStore` on new
`v1alpha1.enrichment_findings` table. OSV + Scorecard ported behind iface w/ unit tests.                                                                              
- **Go client** (`internal/client`) = generic Get/List/Apply/Delete over `v1alpha1.Object`. Legacy per-kind helpers parked in `client_deprecated.go` for transient
callers.                                                                                                                                                              
- **MCP registryserver** ported to v1alpha1 stores. Per-kind services (agent/server/skill/prompt/provider/deployment) + embeddings subsystem deleted.
- **Seed** = v1alpha1 builtin `MCPServer` seeder. Legacy `seed-readme.json` / builtin seed / importer / exporter / embeddings / jobs manager: gone.                   
- **CLI** declarative path (`arctl apply/delete`) rewired to v1alpha1. Imperative `import`/`export`/`embeddings` commands deleted (per-kind CLI rebuild = separate    
branch).                                                                                                                                                              
                                                                                                                                                                    
Net: 219 files, +13,122 / −36,629. KRT reconciler rebase deferred; depends on seams kept here (NOTIFY payload, status-only trigger, `Store.PatchStatus`, soft-delete).
                                                              
# Change Type

/kind breaking_change
/kind cleanup
/kind design                                                                                                                                                          
/kind feature
                                                                                                                                                                    
# Changelog                                                     

```release-note
Ports the registry API, storage, deployment, importer, client, and MCP bridge to a unified Kubernetes-style `ar.dev/v1alpha1` envelope served at `/v0/namespaces/{namespace}/{pluralKind}/{name}/{version}`. Per-kind REST routes, services, and `public.*` tables are removed; existing databases must be recreated.  
```

Additional Notes                                                                                                                                                      
                                                              
- Breaking: wire, routes, DB schema, Go client all change. No dual-stack shim. Operators drop + re-seed registry DB. Legacy public.* migrations gone; new set under   
migrations_v1alpha1/.
- Deferred: KRT reconciler rebase; per-platform deployment identities; Provider credential SecretRef (currently inline in Provider.Spec.Config); more native scanners 
(Trivy, enterprise).                                                                                                                                                  
- CLI: per-kind imperative CLI (publish/pull/list/show) intentionally gone — declarative-only replacement on separate branch.
- Review path: pkg/api/v1alpha1/doc.go (type model) → internal/registry/api/handlers/v0/resource/handler.go (generic handler) →                                       
internal/registry/database/store_v1alpha1.go (store). Commits sequenced as Groups 1–13; each leaves tree green.


EItanya (commented):
We discussed offline, and we're going to use actual schemas for these types rather than JSON blobs

internal/cli/declarative/deployment_apply_test.go
```diff
@@ -14,26 +14,31 @@ import (
...
 spec:
-  resourceType: agent
-  providerId: my-provider
+  targetRef:
+    kind: Agent
+    name: nonexistent-agent
+    version: "0.1.0"
+  providerRef:
+    kind: Provider
+    name: my-provider
```
josh-pritchard, line 28: yeah this was limiting, just ran into it as I couldn't make more than one deployment, nice change

pkg/api/v1alpha1/object.go
```diff
@@ -0,0 +1,84 @@
...
+	// observe it until Finalizers drains, then it is hard-deleted by GC.
+	// Clients MUST NOT set this on apply.
+	DeletionTimestamp *time.Time `json:"deletionTimestamp,omitempty" yaml:"deletionTimestamp,omitempty"`
+
+	// Finalizers are string tokens owned by reconcilers/controllers; each
+	// owner removes its token when its cleanup for this object has finished.
+	// The GC pass requires this slice to be empty before hard-deleting a
+	// terminating row. Clients may add/remove finalizers on apply; the Store
+	// preserves them across spec updates.
+	Finalizers []string `json:"finalizers,omitempty" yaml:"finalizers,omitempty"`
```
yuval-k: not sure we need finalizers, as we are the only controller

pkg/registry/v1alpha1store/store.go
```diff
@@ -0,0 +1,822 @@
...
+		if placeholders != len(opts.ExtraArgs) {
+			return nil, "", fmt.Errorf("%w: fragment references %d distinct placeholder(s) but %d arg(s) supplied",
+				ErrInvalidExtraWhere, placeholders, len(opts.ExtraArgs))
+		}
+		if len(opts.ExtraArgs) > 0 {
+			args = append(args, opts.ExtraArgs...)
+		}
+		if opts.ExtraWhere != "" {
+			where = append(where, rebaseSQLPlaceholders(opts.ExtraWhere, len(args)-len(opts.ExtraArgs)))
+		}
```
yuval-k, line 649: the section here seems to parse and modify an sql query? feel fragine. is it a must have?

xytian315:
Will you please update the sample files here https://github.com/agentregistry-dev/agentregistry/tree/main/examples/declarative

pkg/registry/v1alpha1store/store.go
```diff
@@ -0,0 +1,822 @@
...
+		where = append(where, fmt.Sprintf("labels @> $%d", len(args)))
+	}
+	if opts.Cursor != "" {
+		cursor, err := decodeListCursor(opts.Cursor)
+		if err != nil {
+			return nil, "", err
+		}
+		args = append(args, cursor.UpdatedAt, cursor.Namespace, cursor.Name, cursor.Version)
+		where = append(where, fmt.Sprintf(
+			"(updated_at, namespace, name, version) < ($%d, $%d, $%d, $%d)",
```
yuval-k: the cursor logic here seems coupled to the `ORDER BY` clause below. it's worth adding a comment so the relationship is not missed in future iterations.
also IIUC if an object is updated mid-list, it might appear twice in the final list. so perhaps worth pushing updated_at after namespace and name?

pkg/registry/v1alpha1store/migrations/001_v1alpha1_schema.sql
```diff
@@ -0,0 +1,240 @@
...
+--
+-- Columns:
+--   namespace, name, version      — composite identity (PK)
+--   generation                    — server-managed, bumps on spec mutation
+--   labels                        — user-set key/value, GIN-indexed
+--   spec                          — JSONB per pkg/api/v1alpha1 typed Spec
+--   status                        — JSONB per v1alpha1.Status (ObservedGeneration + Conditions)
+--   is_latest_version             — partial unique index per (namespace, name)
+--   deletion_timestamp            — server-managed soft-delete marker
+--   finalizers                    — reconciler-owned tokens; empty+terminating → GC
```
shashankram, line 82: Why do we need finalizers?

ilackarms: idea was to use them to reconcile orphaned deployments in a follow up

pkg/api/v1alpha1/object.go
ilackarms: idea was to use them to reconcile orphaned deployments in a follow up

pkg/registry/v1alpha1store/migrations/001_v1alpha1_schema.sql
shashankram: I'd like to avoid the k8s style finalizers API that blocks deletion. Can this be a simple field like `phase` that can be set to terminating to signal GC?

josh-pritchard: Either way lets remove  & only include concepts that are used, easy to add things back

pkg/api/v1alpha1/object.go
```diff
@@ -0,0 +1,84 @@
...
+//   - Annotations are narrative: arbitrary key/value pairs for controller
+//     state, tool metadata, etc. Not indexed; can carry larger payloads.
+//     Callers read annotations by key; the server never filters on them.
+//
+// DeletionTimestamp + Finalizers implement Kubernetes-style soft delete:
+// Delete sets DeletionTimestamp and leaves the row in place until every
+// finalizer has been removed by its owner. A GC pass hard-deletes rows
+// where DeletionTimestamp != nil AND Finalizers is empty.
+type ObjectMeta struct {
+	Namespace   string            `json:"namespace" yaml:"namespace"`
```
josh-pritchard: What does namespace mean in AR right now? This is a big concept and that has implications for Auth & policy. Should we bring this in without knowing how it impacts the whole system?

pkg/api/v1alpha1/object.go
```diff
@@ -0,0 +1,84 @@
...
+	Name        string            `json:"name" yaml:"name"`
+	Version     string            `json:"version,omitempty" yaml:"version,omitempty"`
+	Labels      map[string]string `json:"labels,omitempty" yaml:"labels,omitempty"`
+	Annotations map[string]string `json:"annotations,omitempty" yaml:"annotations,omitempty"`
+
+	// Generation is server-managed. Clients MUST NOT set it on apply; the
+	// Store overwrites on every Upsert.
+	Generation int64     `json:"generation,omitempty" yaml:"generation,omitempty"`
```
josh-pritchard: Not sure this has a home here, version makes sense but generation is a global version counter. idk if we want to or need to maintain this.

pkg/api/v1alpha1/registry_validate.go
```diff
@@ -0,0 +1,143 @@
...
+
+// Canonical public registry base URLs the validators check
+// RegistryBaseURL against when the field is set explicitly.
+const (
+	RegistryURLNPM    = "https://registry.npmjs.org"
+	RegistryURLPyPI   = "https://pypi.org"
+	RegistryURLNuGet  = "https://api.nuget.org"
+	RegistryURLGitHub = "https://github.com"
+	RegistryURLGitLab = "https://gitlab.com"
+)
```
josh-pritchard: I don't think constants like this should be in the API if they are not enums like the types above. Users can have private versions of all these so these should be tucked away as defaults somewhere. Maybe they can be here but they should be labeled defaults?

pkg/api/v1alpha1/registries/npm.go
```diff
@@ -47,9 +47,9 @@ func ValidateNPM(ctx context.Context, pkg model.Package, serverName string) erro
 	}
 
 	// Validate that the registry base URL matches NPM exactly
-	if pkg.RegistryBaseURL != model.RegistryURLNPM {
+	if pkg.RegistryBaseURL != v1alpha1.RegistryURLNPM {
 		return fmt.Errorf("registry type and base URL do not match: '%s' is not valid for registry type '%s'. Expected: %s",
-			pkg.RegistryBaseURL, model.RegistryTypeNPM, model.RegistryURLNPM)
+			pkg.RegistryBaseURL, v1alpha1.RegistryTypeNPM, v1alpha1.RegistryURLNPM)
 	}
```
josh-pritchard: I know not changed on this PR but this is a big smell to me. Idk why we even have the RegistryBaseURL in the API if we are just going to validate its an exact match when a user does set it. I figured this would be an override or something but the way this is written RegistryBaseURL should not be a user facing API since we don't allow an override anyways.

internal/registry/api/handlers/v0/deploymentlogs/deploymentlogs.go
```diff
@@ -0,0 +1,106 @@
...
+		out := &deploymentLogsOutput{}
+		out.Body.Lines = []deploymentLogLine{}
+		for line := range ch {
+			out.Body.Lines = append(out.Body.Lines, deploymentLogLine{
+				Timestamp: line.Timestamp.Format("2006-01-02T15:04:05Z07:00"),
+				Stream:    line.Stream,
+				Line:      line.Line,
+			})
+		}
+		return out, nil
```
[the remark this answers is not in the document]
ilackarms: Addressed in 92caf96 — `follow=true` is now rejected with 400 ("not supported on this endpoint; the streaming SSE variant is tracked as a follow-up") and `tailLines=0` is capped at a 10k-line ceiling so a chatty adapter can't blow the response. SSE/chunked variant remains tracked under `REMAINING.md` Group 8.

internal/cli/declarative/resources.go
```diff
@@ -0,0 +1,234 @@
...
+				Version: dep.TargetVersion,
+			},
+			ProviderRef: v1alpha1.ResourceRef{
+				Kind: v1alpha1.KindProvider,
+				Name: dep.ProviderID,
+			},
+			Env:            dep.Env,
+			ProviderConfig: dep.ProviderConfig,
+			PreferRemote:   dep.PreferRemote,
+		},
```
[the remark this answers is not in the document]
ilackarms: Right, the prior shape collapsed distinct deployments. Fixed in 92caf96 — `metadata.{namespace,name,version}` now reflects the deployment row's identity (`dep.Namespace` / `dep.Name` / `dep.Version`); target identity stays in `spec.targetRef`. Two deployments of the same target against different providers round-trip cleanly now.

pkg/registry/resource/readme.go
```diff
@@ -0,0 +1,153 @@
...
+	latestPath := base + "/namespaces/{namespace}/" + plural + "/{name}/readme"
+	versionPath := base + "/namespaces/{namespace}/" + plural + "/{name}/versions/{version}/readme"
+
+	huma.Register(api, huma.Operation{
+		OperationID: "get-latest-" + strings.ToLower(cfg.Kind) + "-readme",
+		Method:      http.MethodGet,
+		Path:        latestPath,
+		Summary:     fmt.Sprintf("Get the latest %s readme", cfg.Kind),
+	}, func(ctx context.Context, in *readmeLatestInput) (*readmeOutput, error) {
+		row, err := cfg.Store.GetLatest(ctx, in.Namespace, in.Name)
```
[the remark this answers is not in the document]
ilackarms: Confirmed via e2e: names containing `/` (DNS-subdomain-style like `e2e-test/decl-mcp-96767`) round-tripped stored unescaped via apply but looked up escaped via path → 404. Fixed in 92caf96 — added an `unescapePath` helper in the generic handler and threaded `url.PathUnescape` through every `{name}` / `{version}` / `{serverName}` path segment in `handler.go`, `readme.go`, and `deployment_logs.go`. 400 returned on malformed escape sequences.

pkg/registry/resource/readme.go
```diff
@@ -0,0 +1,153 @@
...
+		return readmeResponseFromRow(row, cfg.Kind, in.Namespace, in.Name, "", newObj, readmeOf)
+	})
+
+	huma.Register(api, huma.Operation{
+		OperationID: "get-" + strings.ToLower(cfg.Kind) + "-readme",
+		Method:      http.MethodGet,
+		Path:        versionPath,
+		Summary:     fmt.Sprintf("Get a %s readme by namespace, name, and version", cfg.Kind),
+	}, func(ctx context.Context, in *readmeVersionInput) (*readmeOutput, error) {
+		row, err := cfg.Store.Get(ctx, in.Namespace, in.Name, in.Version)
```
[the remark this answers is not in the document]
ilackarms: Confirmed via e2e: names containing `/` (DNS-subdomain-style like `e2e-test/decl-mcp-96767`) round-tripped stored unescaped via apply but looked up escaped via path → 404. Fixed in 92caf96 — added an `unescapePath` helper in the generic handler and threaded `url.PathUnescape` through every `{name}` / `{version}` / `{serverName}` path segment in `handler.go`, `readme.go`, and `deployment_logs.go`. 400 returned on malformed escape sequences.

pkg/registry/resource/readme.go
```diff
@@ -0,0 +1,153 @@
...
+
+	base := strings.TrimRight(basePrefix, "/")
+
+	huma.Register(api, huma.Operation{
+		OperationID: "get-server-readme-v0",
+		Method:      http.MethodGet,
+		Path:        base + "/servers/{serverName}/readme",
+		Summary:     "Get server README",
+	}, func(ctx context.Context, in *legacyServerReadmeLatestInput) (*readmeOutput, error) {
+		row, err := store.GetLatest(ctx, v1alpha1.DefaultNamespace, in.ServerName)
```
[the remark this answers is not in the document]
ilackarms: Confirmed via e2e: names containing `/` (DNS-subdomain-style like `e2e-test/decl-mcp-96767`) round-tripped stored unescaped via apply but looked up escaped via path → 404. Fixed in 92caf96 — added an `unescapePath` helper in the generic handler and threaded `url.PathUnescape` through every `{name}` / `{version}` / `{serverName}` path segment in `handler.go`, `readme.go`, and `deployment_logs.go`. 400 returned on malformed escape sequences.

pkg/registry/resource/readme.go
```diff
@@ -0,0 +1,153 @@
...
+		)
+	})
+
+	huma.Register(api, huma.Operation{
+		OperationID: "get-server-version-readme-v0",
+		Method:      http.MethodGet,
+		Path:        base + "/servers/{serverName}/versions/{version}/readme",
+		Summary:     "Get server README for a version",
+	}, func(ctx context.Context, in *legacyServerReadmeVersionInput) (*readmeOutput, error) {
+		row, err := store.Get(ctx, v1alpha1.DefaultNamespace, in.ServerName, in.Version)
```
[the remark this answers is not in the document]
ilackarms: Confirmed via e2e: names containing `/` (DNS-subdomain-style like `e2e-test/decl-mcp-96767`) round-tripped stored unescaped via apply but looked up escaped via path → 404. Fixed in 92caf96 — added an `unescapePath` helper in the generic handler and threaded `url.PathUnescape` through every `{name}` / `{version}` / `{serverName}` path segment in `handler.go`, `readme.go`, and `deployment_logs.go`. 400 returned on malformed escape sequences.

internal/cli/scheme/scheme.go
```diff
@@ -1,37 +1,110 @@
...
+// back to legacy manifest parsing paths that surface the real error later.
+func IsEnvelopeYAML(data []byte) bool {
+	var header struct {
+		APIVersion string `yaml:"apiVersion"`
+		Kind       string `yaml:"kind"`
+	}
+	if err := yaml.Unmarshal(data, &header); err != nil {
+		return false
+	}
+	return header.APIVersion != "" && header.Kind != ""
```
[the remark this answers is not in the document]
ilackarms: Applied — `IsEnvelopeYAML` now requires `apiVersion == ar.dev/v1alpha1` (alongside the non-empty `kind` check), so generic K8s-style manifests with an unrelated apiVersion don't misclassify into the v1alpha1 decoder.

internal/cli/scheme/scheme.go
```diff
@@ -1,37 +1,110 @@
...
+		obj, ok := item.(v1alpha1.Object)
+		if !ok {
+			return nil, fmt.Errorf("scheme: decoded value does not implement v1alpha1.Object: %T", item)
+		}
+		kindDef, err := reg.Lookup(obj.GetKind())
+		if err != nil {
+			return nil, err
+		}
+		obj.SetStatus(v1alpha1.Status{})
+		out = append(out, resourceFromObject(kindDef.Kind, obj))
```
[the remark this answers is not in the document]
ilackarms: Fixed — `resourceFromObject` now uses `obj.GetKind()` so `Resource.Kind` carries the canonical envelope kind ("Agent", "MCPServer") instead of the registry lookup key. Registry key stays dispatch-only. Tests updated to assert the canonical forms.

internal/cli/skill/pull.go
```diff
@@ -109,7 +120,14 @@ func resolveSkillVersion(skillName, requestedVersion string) (string, error) {
...
 
-	versions, err := apiClient.GetSkillVersions(skillName)
+	versions, err := client.ListVersionsOfName(
+		context.Background(),
+		apiClient,
+		v1alpha1.KindSkill,
+		v1alpha1.DefaultNamespace,
+		skillName,
+		func() *v1alpha1.Skill { return &v1alpha1.Skill{} },
+	)
```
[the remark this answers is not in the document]
ilackarms: Threaded `cmd.Context()` through `resolveSkillVersion` → `ListVersionsOfName`. Ctrl-C and parent timeouts propagate cleanly now.

internal/cli/declarative/deployment_apply_test.go
ilackarms: 👍 thanks!

pkg/api/v1alpha1/object.go
ilackarms: Agreed — converged with @shashankram and @josh-pritchard on the same thread. Hiding from the wire was a halfway step; full removal from the public API is in the next commit on this PR. We'll keep an internal `finalizers` column at the DB layer so the orphan-reconciliation follow-up still has a hook, but `ObjectMeta.Finalizers`, `PatchFinalizers`, `Adapter.ApplyResult.AddFinalizers/RemoveFinalizers`, and the coordinator's finalizer-drain paths are gone from the typed surface.

pkg/registry/v1alpha1store/store.go
ilackarms: Fair — `ListOpts.ExtraWhere` is the seam enterprise uses to inject row-level RBAC filters (see `internal/registry/authz/engine.NewStoreRoleProvider` for the consumer; we can't drop it without losing the authz hook). Added a placeholder-count guard back in `ea68123`; this PR adds a fuzz test for the rebaser plus an explicit contract comment calling out that callers must keep their fragments parameterized (no `fmt.Sprintf`).

pkg/registry/v1alpha1store/store.go
ilackarms: Both fair. Re-ordering to `(namespace, name, version, updated_at)` so the cursor identifies an object rather than a point in time — pagination is snapshot-stable across mid-list updates. Adding an inline comment on `encodeCursor`/`decodeCursor` explaining the format ↔ ORDER BY contract so the next person changing tie-breakers bumps the cursor version field.

pkg/registry/v1alpha1store/migrations/001_v1alpha1_schema.sql
ilackarms: Thread-wide convergence: drop finalizers from the user-facing API entirely. Internal DB column stays for the orphan-reconciliation hook I flagged earlier; `ObjectMeta.Finalizers`, `PatchFinalizers`, and `Adapter.AddFinalizers/RemoveFinalizers` go away. Soft-delete via `deletion_timestamp` survives. Shipping in the next commit on this PR; release note added.

pkg/api/v1alpha1/object.go
ilackarms: Addressed in fe3b126. Namespace stays in the data model (storage identity is `(ns, name, version)`) but is hidden from the wire: defaults to `"default"`, stripped from JSON responses when default, only surfaceable via `?namespace=...` query when explicitly opted in. `?namespace=all` is the cross-namespace escape hatch on list endpoints. Multi-namespace remains a future, intentional flip rather than something accidentally exposed today.

pkg/api/v1alpha1/object.go
ilackarms: Hidden from the wire in 7c905fc (`json:"-"`). It stays internal because the Phase 2 KRT reconciler uses it as the convergence signal (`status.observedGeneration` vs `metadata.generation`). Users never see or set it. If KRT picks a different convergence marker we can drop the field outright; until then this is the simplest pairing.

pkg/api/v1alpha1/registry_validate.go
ilackarms: Agreed. The well-known `RegistryType*` enums are protocol and stay in the API package, but the per-registry default URLs (`registry.npmjs.org`, etc.) are policy not protocol. Moving them to a `pkg/api/v1alpha1/registries/defaults` subpackage and renaming to `Default<X>RegistryURL` so they read as overrideable defaults rather than canonical constants. Shipping in the next commit on this PR.

pkg/api/v1alpha1/registries/npm.go
ilackarms: Agreed — current "must equal the canonical URL" behavior is the worst of both. Going with option 2: when `RegistryBaseURL` is set on a package, the validator uses it; when empty, the registry-type default. Lets enterprise users point at private NPM mirrors / private OCI registries without hacking the validator. Shipping in the next commit on this PR.

josh-pritchard (approved):
LGTM! Tested locally

PR #451, opened 2026-04-21
agent: add version to AgentMcpServer to allow name+version ref

shashankram:
# Description
A recent change regressed the previous behavior that allowed name+version refers in `mcpServers`. This fixes it.

# Change Type

```
/kind bug_fix
```

# Changelog

```release-note
NONE
```


xytian315 (approved):
LGTM

PR #454, opened 2026-04-23
--force flag for delete deployment

josh-pritchard:
# Description

Add `--force` flag to `arctl delete deployment` that skips provider-specific teardown and only removes the deployment record from the registry.

**Motivation:** When a provider returns errors during deletion (e.g., the resource was already deleted out-of-band, or the provider returns authorization errors), the deployment record gets stuck in AR with no way to remove it. This provides an escape hatch for recovery.

**What changed:**
- `arctl delete deployment <name> --version <version> --force` skips the adapter `Undeploy` call and goes straight to record removal
- `DELETE /v0/deployments/{id}?force=true` query param added to the API
- `DeploymentService.UndeployDeployment` accepts a `force bool` parameter — when true, the authz check still runs but the adapter teardown is skipped
- `Client.DeleteDeployment` accepts `force bool` and appends the query param

# Change Type

/kind feature

# Changelog

```release-note
feat: add `--force` flag to `arctl delete deployment` to skip provider teardown and remove orphaned deployment records
```

# Additional Notes

This is a breaking change to the `DeploymentService` interface (`UndeployDeployment` signature gains a `force bool` param). Enterprise callers will need to update accordingly.


xytian315 (approved):
Looks good to me for interim solution!

Thanks so much!

PR #462, opened 2026-04-30
Remove public semantic search surface until we can rework it

iplay88keys:
# Description
Removes the public surface of semantic search until it can be reworked as there are issues with it at this point so we don't have a broken feature. There is also missing e2e test coverage around it at this point.

Included:
- Remove Helm value (`vectorEnabled`)
- Make sure vector migrations are disabled -- previously keyed off of `vectorEnabled` helm value
- Skip registering indexing routes
- Don't output semantic search info on list endpoints (generic)
- Disable `pgvector` image config in Makefile
- Update embedding migration to use `IF NOT EXISTS` for idempotency if we decide to split out into a separate track in the future.
- Remove configuration from `internal/registry/config/config.go`

# Change Type

```
/kind cleanup
```

# Changelog
```release-note
Remove semantic search and embeddings
```

# Additional Notes
- https://github.com/agentregistry-dev/agentregistry/issues/217


timflannagan (approved):
LGTM. One quick thing: internal/registry/config/config.go. I think we should rip out everything besides the enabled field and add a TODO comment for implementing semantic search. WYDT?

timflannagan (approved):

iplay88keys:
>LGTM. One quick thing: internal/registry/config/config.go. I think we should rip out everything besides the enabled field and add a TODO comment for implementing semantic search. WYDT?

I think that's for the best so someone doesn't see that in documentation and be confused when setting those values doesn't work.

PR #464, opened 2026-04-30
Move telemetryEndpoint from AgentSpec to ProviderSpec

josh-pritchard:
# Description
Telemetry endpoint is runtime config, not part of an agent's identity. Moving it onto ProviderSpec lets the same Agent be deployed to different environments with different OTLP collectors.

Also actually wire this up at deploy time since it wasn't passed into agent runtime yet & change to use OTEL_EXPORTER_OTLP_ENDPOINT so it automatically just works with endpoint:port and users don't need to know to use `v1/traces`

# Change Type
/kind cleanup

# Changelog

```release-note
Move telemetryEndpoint from AgentSpec to ProviderSpec
```

# Additional Notes


internal/cli/agent/frameworks/adk/python/templates/docker-compose.yaml.tmpl
```diff
@@ -2,18 +2,16 @@
 # Generated by the AgentRegistry CLI.
 
 services:
-{{- if .TelemetryEndpoint}}
```
josh-pritchard, line 5: was kind of nonsense that this was a flag that could point it anywhere but it also just ran a local one then for debug?

timflannagan (commented):
Should this live in the Provider API instead of the Deployment API?

timflannagan (approved):

PR #476, opened 2026-05-06
breaking change: makes versioning immutable for content-registry kinds (Agent, MCPServer, RemoteMCPServer, Skill, Prompt)

xytian315:
# Description

**Motivation:** 
Makes versioning immutable for content-registry kinds (Agent, MCPServer, RemoteMCPServer, Skill, Prompt). Versions are system-assigned monotonic integers, stored in status.version. apply is append-only — a changed spec produces a new version row instead of mutating in place.

**What changed:** 
Adds:

GET /v0/{plural}/{name}/versions — list all versions of a content-registry resource
arctl get/delete --version and --all-versions flags
Auditor interface in pkg/types so consumers can record state-change events at the source
# Change Type

**Breaking changes:**
  - `metadata.version` is rejected on input for content-registry kinds.                                                                                             
  - Direct `PUT /v0/{plural}/{name}/{version}` removed for content-registry kinds (use `POST /v0/apply`).                                                         
  - DB schema for content-registry tables: `version` is now `INTEGER`; `is_latest_version`, `generation`, `finalizers` columns dropped; `spec_hash` added.          
  - Re-applying the same name with a changed spec creates a new version row (was: mutates in place).                                                                
  - Go API: `Status.ObservedGeneration`, `Metadata.Generation`, `Condition.ObservedGeneration` removed.  

/kind feature
/kind breaking_change
# Changelog

```release-note
breaking: Content-registry kinds (Agent, MCPServer, RemoteMCPServer, Skill, Prompt) now use immutable system-assigned integer versioning. `metadata.version` is rejected on input; `status.version` carries the assigned integer. `apply` is append-only — a changed spec creates a new version row.
```

# Additional Notes


ilackarms:
Deep review note on immutable versioning / downstream integration:

1. **Version numbers should not be reused after all-version delete.**

Right now `DeleteAllVersions` frees the `(namespace, name)` identity and the next apply starts at version 1 again. That makes `Agent/foo@1` reusable for a different spec after delete/recreate, which breaks the meaning of pinned refs, audit records, and deployments that reference concrete versions. If versions are server-assigned immutable identities, the allocator needs durable per `(namespace, name)` counter state or tombstones so version numbers are never reused, even after all content rows are deleted.

2. **Please preserve/expose a version planning seam before production mutation.**

This PR moves the new-version decision entirely inside `Store.Upsert`, which makes it hard for enterprise approval/admission to stage changed-spec writes before production mutation. Approval needs to know whether an apply is a no-op, metadata-only update, legacy mutable update, or new immutable version, plus the reserved version/spec hash, before the production row is inserted and before post-upsert hooks run. The enterprise approval PR can consume that seam, but this PR should avoid making production `Upsert` the only place where version math can happen.

3. **Downstream v1alpha1 kinds need an explicit store-mode contract.**

Built-in Provider/Deployment use `NewDeploymentStore` legacy mode, but extra stores from `AppOptions.V1Alpha1StoreTables` are always constructed with `NewStore`, which now means versioned-artifact/integer/spec_hash mode. Enterprise AccessPolicy is a v1alpha1 envelope kind but remains mutable/legacy-shaped (`VARCHAR version`, generation/finalizers/is_latest_version), so this wiring will register the wrong routes and issue SQL against missing columns. Please expose a way for downstream kinds to opt into legacy/mutable storage, or make the extra-store registration carry mode metadata.

xytian315:
> feat(v1alpha1store): split Upsert into Plan and Apply for approval st…

@ilackarms Could you please check recent commits? I think it should address all your feedback.

timflannagan:
Superseded by #482.

PR #483, opened 2026-05-08
rename provider->runtime and update runtime types

shashankram:
# Description

Renames the v1alpha1 `Provider` resource kind to `Runtime` to match the agentic-runtime terminology used by hosted vendors (Bedrock AgentCore, Gemini Agent Runtime). The discriminator field becomes `spec.type` - free-form, case-insensitive - replacing `spec.platform`. Built-in runtime types are `Local` and `Kubernetes`; downstream builds register additional canonical CamelCase values via the existing extension hook.

The deployment surface follows: `spec.providerRef`/`spec.providerConfig` become `spec.runtimeRef`/`spec.runtimeConfig`, and the CLI exposes `runtime`/`runtimes` instead of `provider`/`providers`. The HTTP API moves from `/v0/providers/...` to `/v0/runtimes/...`, the storage table from `v1alpha1.providers` to `v1alpha1.runtimes`, and adapter annotations from the `platforms.agentregistry.solo.io/...` prefix to `runtimes.agentregistry.solo.io/...`.

The deployment-adapter interface is renamed accordingly (Type, ApplyRuntime, RemoveRuntime, Runtime field on Apply/Remove/Discover inputs); built-in adapters dispatch via case-insensitive lookup so any casing of `spec.type` routes correctly. The OpenAPI spec is regenerated, the embedded UI follows the same field rename, and authz permission artifact types pick up the new `runtime` scope.

Old manifests (`kind: Provider`, `spec.platform`, `spec.providerRef`) are rejected.

# Change Type

```
/kind breaking_change
/kind cleanup
```

# Changelog

```release-note
NONE
```


internal/registry/runtimes/local/adapter.go
```diff
@@ -16,29 +16,29 @@ import (
...
+func NewLocalDeploymentAdapter(runtimeDir string, agentGatewayPort uint16) *localDeploymentAdapter {
 	return &localDeploymentAdapter{
-		platformDir:      platformDir,
+		runtimeDir:       runtimeDir,
 		agentGatewayPort: agentGatewayPort,
 	}
 }
 
-func (a *localDeploymentAdapter) Platform() string { return "local" }
+func (a *localDeploymentAdapter) Type() string { return v1alpha1.TypeLocal }
```
xytian315, line 41: small thing — `Type()` here returns `"Local"` but the adapter map in `registry_app.go` is keyed by `"local"` lowercase. works because lookups `resolveAdapter` do `strings.ToLower`, but `m[adapter.Type()] = adapter` would silently miss.

ilackarms (approved):

PR #488, opened 2026-05-12
Add generic apply admission seam for approval flows

ilackarms:
# Description

Adds a neutral OSS apply admission seam that downstream builds can use to accept a validated apply before production storage side effects. The seam runs after auth, object validation, reference validation, and registry validation, but before `Store.Upsert` and `PostUpsert`.

Key pieces:
- Adds `ApplyInterceptor` and `ApplyObject` so downstream approval replay can reuse the same production apply path.
- Exposes finalized v1alpha1 route context through `AppOptions.ExtraResourceRoutes`.
- Adds an optional resolver wrapper and import authorizer override for downstream integrations.
- Changes Deployment reconciliation so dangling target/runtime refs persist a `Ready=False` / `ReferencePending` status instead of failing the stored Deployment apply.

This intentionally avoids enterprise approval names and does not reintroduce public `metadata.version`, `targetRef.version`, deployment version identity, or integer-version compatibility.

# Change Type

/kind feature

# Changelog

```release-note
Added a generic apply admission seam for downstream integrations and pending-reference deployment status handling.
```

# Additional Notes

Stack base for the enterprise native approval flow PR.

KRT note: `ApplyInterceptor`, `ResolverWrapper`, and `ExtraResourceRoutes` are marked in code as temporary synchronous-handler bridges. They should be removed or collapsed when KRT/reconciler-owned admission and staging becomes the production write architecture.

Verified:
- `rtk go test -tags=integration ./pkg/registry/resource ./internal/registry/service/deployment ./internal/registry/api/router ./internal/registry`
- `rtk go test ./pkg/api/v0 ./pkg/types ./pkg/registry/resource ./internal/registry/service/deployment ./internal/registry/api/router ./internal/registry`
- `rtk go test ./pkg/types ./pkg/registry/resource ./internal/registry/api/router`


internal/registry/service/deployment/coordinator.go
```diff
@@ -95,10 +95,16 @@ func (c *Coordinator) Apply(ctx context.Context, deployment *v1alpha1.Deployment
 
 	target, err := c.resolveTarget(ctx, deployment)
 	if err != nil {
+		if errors.Is(err, v1alpha1.ErrDanglingRef) {
+			return c.persistReferencePending(ctx, deployment, err)
+		}
```
xytian315, line 100: Question: What happens here if targetRef/runtimeRef is genuinely missing (typo / deleted resource)? Should we split into ErrPendingRef vs ErrDanglingRef, so a typo'd ref still fails loudly?

ilackarms: typo still fails earlier: `applyCore` runs `ResolveObjectRefs` before persistence, so bad target/runtime ref does not write Deployment row

timflannagan (approved):

PR #491, opened 2026-05-13
Remove import enrichment pipeline

timflannagan:
# Description

Removes the legacy import enrichment path from upstream. The canonical
v1alpha1 resource API remains `/v0/apply` plus the generic `/v0/{plural}`
resource routes; this only drops `/v0/import`, the importer package, and the
Scorecard/OSV scanner implementations that were only reachable through that
pipeline.

This also removes the stale import/score UI affordances, keeps the old
findings migration versions as no-op placeholders, and tidies the Go module so
Scorecard, OSV scanner, Tablewriter, and their transitive dependencies fall out
of the build graph.

# Change Type

/kind breaking_change

# Changelog

```release-note
Removed the legacy /v0/import enrichment endpoint and scanner pipeline.
```


pkg/registry/v1alpha1store/migrations/002_enrichment_findings.sql
inFocus7, line 1: Q: Shouldn't we create a new migration file to drop this table instead of editing this pre-existing one so that upgrades drop it correctly? (`DROP TABLE IF EXISTS v1alpha1.versionedTable`)

pkg/registry/v1alpha1store/migrations/006_enrichment_findings_tag.sql
inFocus7, line 1: Q: maybe the same question as `pkg/registry/v1alpha1store/migrations/002_enrichment_findings.sql`, although we'd just keep this file as-is and let the future `DROP` migration handle the table removal.

inFocus7 (changes requested):
overall good, although we'll want to revert the changes on the two sql files and create a new versioned migration file that `DROP`s the table. Else, upgrades will keep the now-unused tables.

edit: local spin-up UI lgtm (nothing broken at least)

inFocus7 (approved):
lgtm - i just pushed an update to create a separate migration file for dropping the enrichments table

PR #498, opened 2026-05-15
fix(cli): arctl help offline

xytian315:
# Description
- **`arctl help` works offline** — cobra's auto-generated `help` subcommand
    was going through the root `PersistentPreRunE`, which calls `preRunSetup`
    and pings the registry. With no registry running, `arctl help` and
    `arctl help <command>` failed with `registry unreachable`, while
    `arctl --help` worked (cobra handles the flag before run hooks). Added
    `help` to `preRunSkipCommands["arctl"]`, matching the existing pattern
    for `configure` / `completion` / `init` / `build`. All three help paths
    are now offline.
 - **adk-python compose OTLP host ports** — bump the otel-collector host-side
    mappings to `14317`/`14318` so `arctl init agent` doesn't generate a
    compose file that collides with other OTLP collectors that conventionally
    bind `4317`/`4318` on the host

# Change Type

/kind fix
# Changelog

```release-note
NONE
```

# Additional Notes

### DIFF
--- a/internal/cli/frameworks/builtin/adk-python/templates/docker-compose.yaml.tmpl
+++ b/internal/cli/frameworks/builtin/adk-python/templates/docker-compose.yaml.tmpl
@@ -4,9 +4,16 @@
 services:
   otel-collector:
     image: otel/opentelemetry-collector:latest
+    # Host-side ports are bumped to 14317/14318 so they don't collide with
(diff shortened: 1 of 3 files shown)


timflannagan (approved):
Fine for now obviously, will need some cleanup here in the future, etc.

josh-pritchard (approved):

PR #500, opened 2026-05-15
Remove semantic search + embeddings

inFocus7:
Built on top of other cleanup work #491 

note: this got autoclosed since the base branch merged, moved changes to #505 

# Description

Removes the unwired semantic-search / embeddings feature. The HTTP handlers, the indexer, and the public Store surface are deleted; the `semantic_embedding*` columns and HNSW indexes are dropped by a new v1alpha1 migration on upgrade.

What's removed:
- `internal/registry/api/handlers/v0/embeddings/`: the indexer kickoff and job-status endpoints. Never wired into the running router.
- `internal/registry/embeddings/`: the indexer, helpers, and Provider interface. Never constructed at boot.
- ⚠️ `pkg/semantic/` and the `v1alpha1store.Store` semantic methods (`SetEmbedding`, `GetEmbeddingMetadata`, `SemanticList`, `VectorLiteral`). **Public-API breaking**: downstream consumers importing these symbols will get compile errors.
- `EmbeddingsConfig` and the `AGENT_REGISTRY_EMBEDDINGS_ENABLED` env var.
- The migrator's generic `MigratorConfig.Skip` predicate, which only ever gated 003.

Migrations:
- `003_embeddings.sql` is retained as a no-op (comment-only) for historical consistency with installs that have version 203 recorded in `schema_migrations`. Plain postgres images can apply it.
- `008_drop_semantic_embeddings.sql` drops the columns and HNSW indexes with `IF EXISTS`, safe on installs that never ran 003 originally.

# Change Type

```
/kind breaking_change
/kind cleanup
```

# Changelog

```release-note
Remove embeddings and semantic search.
⚠️: If you enabled embeddings and use the pgvector postgres image, keep the image during the upgrade. You can switch to a standard postgres image in a follow-up release.
```

# Additional Notes

## Upgrade notes

Operators currently running with `AGENT_REGISTRY_EMBEDDINGS_ENABLED=true` on a postgres image that includes pgvector (e.g. `pgvector/pgvector:pg16`) must keep the pgvector image during this upgrade.

Migration 208 drops the semantic_embedding* columns and HNSW indexes, which are typed against pgvector. PostgreSQL requires the extension's shared library to be loadable on disk to execute the drops. Attempting to upgrade the application and switch to a plain postgres image in _the same_ Helm release fails with:

```
ERROR: could not access file "$libdir/vector": No such file or directory
```

To remove the pgvector dependency:
1. Upgrade the application first, keeping the pgvector image.
2. Verify migration 208 applied: SELECT version FROM schema_migrations WHERE version=208;
3. In a follow-up Helm release, swap the bundled postgres image to plain postgres of the same major version.

Operators who ran with EMBEDDINGS_ENABLED=false (the default) have nothing to drop and can upgrade freely on plain postgres.


pkg/registry/v1alpha1store/migrations/003_embeddings.sql
inFocus7, line 1: we either need to delete this file or no-op it, since we no longer implement embeddings or its config (e.g. embeddings enablement), meaning we won't do in-code `Skip`. because of that, without a skip we would be enabling the extension, which doesn't exist in base postgres.

on fresh installs this migration will apply, but not do anything, keeping a consistent 00N list of migrations. On upgrades, this file isn't re-applied, so no issues there.

PR #502, opened 2026-05-18
Dev tooling updates

inFocus7:
# Description

Updates to the dev workflow

1. Move tool dependencies to own `tools/` folder to separate from root dependencies.
2. Added Helm doc file generation via `make charts-docs`
    - There was a pre-existing `README.md.gotmpl` file, but it wasn't wired up to generate the `README.md`.
4. Updated `lint` workflow step to use go tool instead of GH action
5. Added `tools/go.sum` to dependency caches in workflows
6. Using `gci` within `golanci-lint`**\***

**\*** A lot of go imports where refactored due to the `gci` settings:
```yaml
formatters:
  settings:
    gci:
      sections:
        - standard
        - default
        - localmodule
```

# Change Type

```
/kind cleanup
```

# Changelog

```release-note
NONE
```

# Additional Notes

Verification based on workflows running successfully.

Another verification is simply running one of the tool-based make targets - such as `make lint` - and see it work.


Makefile
```diff
@@ -45,9 +45,20 @@ LDFLAGS := \
 # Local architecture detection to build for the current platform
 LOCALARCH ?= $(shell uname -m | sed 's/x86_64/amd64/' | sed 's/aarch64/arm64/')
 
+# Developer tools (golangci-lint, gotestsum, gci, kind) are pinned in tools/go.mod
+# and invoked via `go tool -modfile=tools/go.mod <name>`. This keeps their
+# transitive dependencies out of the main module.
+TOOLS_DIR     ?= $(CURDIR)/tools
+GO_TOOL       ?= go tool -modfile=$(TOOLS_DIR)/go.mod
+CRANE         ?= $(GO_TOOL) crane
+GCI           ?= $(GO_TOOL) gci
```
timflannagan: Do we need either of these tool deps?

timflannagan: We should be using the gci golangci-lint formatter instead of the direct tool dependency. That can be done in a follow-up though. You can see the commented out config already for enabling it.

timflannagan (commented):
QQ on crane

Makefile
inFocus7: Good calls, `crane` was unused, but I initially kept it (in case). I just removed it and the direct call to `gci` + `crane` usage.

on a relevant note: `gci` was originally called only modified files, but since it's now part of the `golangci-lint` flow it runs for all files. Some go import formatting came into my changes 👍🏼

timflannagan (approved):

PR #503, opened 2026-05-19
feat: swap to golang-migrate v4; consolidate migrations + parameterize schema

nikolasmatt:
# Description

> **Upgrade-safe, with a documented downgrade floor.** This PR replaces the hand-rolled migrator with `golang-migrate/migrate v4`, collapses every prior OSS migration (`001..008`) into a single `001_initial_schema`, and moves the schema name from the hard-coded `v1alpha1` to a parameterized `agentregistry` (default; future work will make it env-driven). Existing deployments bridge transparently on first contact via a one-shot data-copy from `v1alpha1.*` into the new schema. **Once bridged, downgrading to a pre-engine-swap binary is not supported.**[^downgrade-floor]

Adds `arctl db migrate` so operators can apply, roll back, and inspect Postgres migrations independently of server startup, and surfaces `SkipMigrations` for booting the server against an already-migrated schema.

Operator surface:

```
arctl db migrate up                apply pending migrations
arctl db migrate down N            roll back N most-recent
arctl db migrate status            applied / pending counts
arctl db migrate version           highest applied version
arctl db migrate goto V            move schema to version V (V=0 empties)
arctl db migrate force V           mark V applied without running its SQL
```

Per-source ops (`down`, `goto`, `force`, `version`) take `--source X` when more than one source is registered; inferred for OSS-only installs. `up` always runs every registered source through the orchestrator and does not accept `--source`. DSN via `--db-url` or `AGENT_REGISTRY_DATABASE_URL`.

Server-side, `AppOptions.SkipMigrations` (or the `SKIP_MIGRATIONS` env var — deliberately prefix-free, unlike the `AGENT_REGISTRY_`-prefixed config, so the same name toggles the gate across binaries) makes the server skip the OSS startup migrator. The pool is still opened and pinged so a misconfigured DB fails fast.

Key decisions:

- **Engine: `golang-migrate/migrate v4` + pgx/v5 driver + iofs source.** Replaces the hand-rolled migrator. `search_path` is injected into the migrator's DSN so unqualified migration DDL lands in the target schema regardless of the connecting user (the `SchemaName` config alone does not set it). Dirty-state recovery is not automated — the idempotent-DDL convention plus the advisory lock are the production contract; operators clear a dirty marker after a failed `down` via `arctl db migrate force V`.

- **Schema as a resolved value, not a string.** Each source owns its own Postgres schema, modeled as a `database.Schema` (a validated name with its quoted identifier precomputed) resolved once and threaded through the migrator, the orchestrator, and the stores — never re-sanitized per call. The migrator's `schema_migrations` bookkeeping table lives in that schema, so two sources sharing a database keep their counters separate without colliding. Default `agentregistry` (was hard-coded `v1alpha1`); a `SchemaRegistry` resolves source-name → schema. Schema names are lowercase-only, so the quoted and unquoted forms always refer to the same schema.

- **Explicit qualification, not `search_path` reliance.** App queries name their schema explicitly (`"agentregistry"."agents"`) via the injected `Schema`, so a store stays correct even on a connection whose `search_path` points at a different schema. The pool's `AfterConnect` still sets `search_path` to the OSS schema as a default.

- **Consolidated baseline + reusable lint.** Every prior OSS migration (001..008) collapses into a single schema-agnostic `001_initial_schema.up.sql` derived from `pg_dump`. The lint is a standalone, exported package (`pkg/registry/database/migrationlint`); any migration set runs the same `Check(fs, dir)` against its own embed. It rejects schema-qualified identifiers across the DDL/DML surface (CREATE/ALTER/DROP of table/index/view/materialized view/sequence/type/trigger/function, INSERT/UPDATE/TRUNCATE, REFERENCES, FROM/JOIN, COMMENT ON, GRANT/REVOKE — bare and double-quoted) plus the constructs that break single-file atomicity (CONCURRENTLY, explicit transaction control, and `VACUUM` / `CREATE|DROP DATABASE|TABLESPACE` / `ALTER SYSTEM`).

- **Idempotent legacy bridge as plain Go.** `orchestrator.RunUp` applies `Steps(1)` to install the new schema, invokes an optional `LegacyRun` callback (gated on `public.schema_migrations` existing), then runs `Up()`. The OSS `LegacyRun` copies rows from `v1alpha1.*` into `agentregistry.*` via column-enumerated `INSERT … ON CONFLICT DO NOTHING` inside a single transaction; each `v1alpha1.<t>` is probed individually so a manually-dropped legacy table doesn't fault the whole copy. After any source's bridge actually fires, `public.schema_migrations` is renamed to `public.schema_migrations_v0_legacy` (gated on the bridge so an unrelated owner of that well-known table name is never silently touched). The bridge survives a partial-run abort (Steps(1) committed, process died before `LegacyRun`) because the gate is the legacy-table existence and the copy is idempotent. No operator action required — except one rejected case: a deployment that never finished the pre-engine-swap migration sequence is refused up front with an actionable error (the column-enumerated copy assumes the final pre-engine-swap schema, so a short upgrade would otherwise abort mid-copy or skip later-added data; the operator finishes the prior release's migrations and re-runs).

- **Serialized, all-or-nothing startup.** A single global `pg_advisory_lock` is held across the entire `RunUp` (every source plus the failure-restore path), so concurrent pods run the sequence strictly one at a time. Per-source locks additionally serialize orchestrator `up` against per-source CLI ops (`down`/`goto`/`force`), acquired by the CLI via `orchestrator.WithSourceLock`. If any source fails, every source touched this run — including the failing one — is restored to the version it sat at before the run, so the database returns to the prior release's version combo rather than a cross-track state no release ships; a source found already dirty at entry is surfaced for `force` rather than guessed. This is why the lint enforces single-file atomicity: a failed migration file must roll back its DDL, leaving only a dirty marker for the restore to clear.

- **Up-only convention.** `NNN_name.up.sql` + `NNN_name.down.sql` pair (iofs requires the pair). The single OSS migration ships a `RAISE EXCEPTION 'not reversible'` down. Idempotent DDL (`CREATE … IF NOT EXISTS`, `CREATE OR REPLACE`) makes retry of a partially-applied migration safe.

- **CLI safety rails.** `force V` rejects unshipped versions (no typo-wedging the DB via `force 9999`). `down N` reports the actual rolled-back count from the pre/post version delta, not the user-supplied N. `version` and `goto V` disambiguate "(no migrations applied)" / "(dirty)". `down`'s help names `force V` as the recovery path for up-only migrations. `status -o json` is a frozen wire contract locked by a golden test.

- **Extension hooks.** `orchestrator.Source{Name, Schema, Files, Dir, LegacyRun}` registers an additional migration source (`Schema` is a `database.Schema`; `LegacyRun` receives the resolved schema). Compose by passing a `Source` to `RunUp`, or by calling `migrate.Register` in a CLI's `init()` to pick up the per-source subcommands. `migrationlint`, `SchemaRegistry`, and the `Schema` type are exported so a downstream migration set runs the identical lint and resolves its schema the same way. An extra v1alpha1 store kind registered via `AppOptions.V1Alpha1StoreTables` as `"schema.table"` resolves into that schema; a bare `"table"` stays in the OSS schema.

The old `v1alpha1.*` tables are left in place by the bridge; a future release will ship a regular migration that drops them.

Out of scope (called out for follow-up): the residue-drop migration; env-driven schema-name configurability; Helm pre-install Job.

# Change Type

/kind feature

# Changelog

```release-note
Added `arctl db migrate` (up, down N, status, version, goto V, force V) for applying, rolling back, and inspecting Postgres migrations independently of server startup. The OSS migrator now uses `golang-migrate/migrate v4` with the schema name parameterized; tables live in the `agentregistry` schema (was `v1alpha1`). Existing deployments are bridged transparently: data copies from `v1alpha1.*` into `agentregistry.*` on first contact and `public.schema_migrations` is renamed to `public.schema_migrations_v0_legacy`. Downgrading to a pre-engine-swap binary is not supported once bridged. Adds `AppOptions.SkipMigrations` / the `SKIP_MIGRATIONS` env var to boot the server against a pre-migrated schema.
```

# Additional Notes

`make verify` is clean. Integration + e2e tiers run against a local Postgres:

- **Orchestrator integration** (`pkg/registry/database/orchestrator/orchestrator_test.go`, `integration` tag) — fresh install, idempotency, legacy-bridge data preservation (`v1alpha1.*` → `agentregistry.*` with byte-identical payloads, `public.schema_migrations` rename), bridge re-run idempotency, the partial-run-abort recovery path, a 5-goroutine multi-pod race against the advisory lock, and the failure-restore paths (prior sources rolled back, an upgraded failing source restored to its entry version, a fresh-install failure reset, and an already-dirty source refused).

- **Legacy bridge** (`pkg/registry/database/legacymigrate/oss_integration_test.go`, `integration` tag) — a below-floor OSS track is refused before any data is copied, and a track exactly at the last pre-engine-swap release bridges normally.

- **Schema + search_path** — `pkg/registry/database/schema_test.go` (name validation, quoting, `Qualify`, registry) and a `search_path` regression test (`migrate_searchpath_test.go`) asserting unqualified DDL lands in the configured schema even when it differs from the connecting user.

- **Migration lint** (`pkg/registry/database/migrationlint`) — the OSS set runs `migrationlint.Check` on every `go test`; the package's own tests exercise each forbidden-pattern category (positive + negative, bare + quoted) and confirm PL/pgSQL block syntax and unqualified statements stay clean.

- **CLI e2e** (`pkg/cli/db/migrate/cli_e2e_test.go`, `e2e` tag) — execs the arctl binary as a subprocess. Covers missing-DSN, `--db-url` precedence, `--help`, arg validation, happy path, idempotent re-`up`, down-not-reversible, `force` writes the bookkeeping row only, `force` rejects out-of-range V, fresh-DB `version` disambiguation, JSON status output and its frozen shape, and `--source` inference.

Migration-authoring docs: `pkg/registry/v1alpha1store/migrations/README.md` covers the schema-agnostic + single-transaction conventions, the lint contract, and the downgrade floor.

[^downgrade-floor]: AgentRegistry has not supported downgrades historically; this release sets a documented horizon rather than regressing a previously-supported contract.


pkg/cli/db/migrate/migrate.go
```diff
@@ -0,0 +1,437 @@
...
+					return nil
+				}
+				fmt.Fprintf(cmd.OutOrStdout(), "applied %d migration(s); schema is up to date\n", total)
+				return nil
+			})
+		},
+	}
+}
+
+func newDownCmd() *cobra.Command {
```
iplay88keys, line 303: My biggest concern with this is that there is a singular version here, yet the OSS migrations and enterprise migrations are stacked. So if `2026.05.01` has OSS 201->205 and Ent 501->507. Then we upgrade to `2026.06.01` which has OSS 201->206 and Ent 501->508, we can only downgrade the Enterprise migrations and have no way to get the entire db back in the `2026.05.01` state

dhaifley:
I think this approach, and UX in generally sound. As we discussed during our meeting today, I think packaging the migrations into a CLI tool makes the most sense, and using a tool at least as new as the highest version for any upgrade or downgrade to apply the migrations.

I think the interface this exposes can be used universally across projects. kagent-enterprise does not currently have a CLI tool, but can build a dbctl tool that can perform a subset of the functions described by this, and also can perform database snapshots and restores. It also applies to ClickHouse rather than PostgresSQL. It would be possible to extend that to support the exact interface used by arctl in this PR, including version based migrations and goto version, and adding a force command. If/when a CLI tool for kagent-enterprise exists, this dbctl functionality could be included as db commands within it, in this same way as arctl does in this PR.

For enterprise, having a single utility that can process ClickHouse and Postgres would be possible if multiple connection strings can be provided via multiple flags or env vars.

On the enterprise side, the dbctl tool is built from a library, so the migrations can also be applied at startup for the ui-backend service. That is the typical way they would be applied. I think that would continue to be possible if this interface was fully implemented.

I think you have covered everything needed in this PR in terms of the UX, and the same can be applied to enterprise. I can't comment on anything AR specific, since I don't have much context, but this approach and interface for db migrations looks good to me, and we are using similar approaches in enterprise, and other places, which could be adjusted to be the exact same.

pkg/cli/db/migrate/migrate.go
iplay88keys, line 1: One thing I noticed is there is no dirty flag, so partial migrations would leave the db in an inconsistent state.

pkg/cli/db/migrate/migrate.go
```diff
@@ -0,0 +1,437 @@
...
+				for _, src := range srcsList {
+					m := database.NewMigrator(conn, src.cfg)
+					applied, pending, err := m.Status(ctx)
+					if err != nil {
+						return err
+					}
+					appliedTotal += len(applied)
+					pendingTotal += len(pending)
+				}
+				fmt.Fprintf(cmd.OutOrStdout(), "%d migration(s) applied, %d pending\n", appliedTotal, pendingTotal)
```
iplay88keys: It might be useful to have this info available as json, but definitely not necessary.

pkg/cli/root.go
```diff
@@ -139,6 +144,34 @@ func init() {
...
+	migrate.Register(migrate.Source{
+		Name: "oss",
+		BuildConfig: func() database.MigratorConfig {
+			enabled, set := migrate.EmbeddingsEnabledOverride()
+			if !set {
+				if raw := os.Getenv("AGENT_REGISTRY_EMBEDDINGS_ENABLED"); raw != "" {
+					parsed, err := strconv.ParseBool(raw)
+					if err != nil {
+						fmt.Fprintf(os.Stderr, "failed to parse AGENT_REGISTRY_EMBEDDINGS_ENABLED=%q: %v\n", raw, err)
+						os.Exit(1)
```
iplay88keys: This would block `arctl db migrate --help` as well, but isn't a big problem.

internal/registry/config/config.go
```diff
@@ -34,6 +35,15 @@ type Config struct {
 
 	// Embeddings / Semantic Search
 	Embeddings EmbeddingsConfig
+
+	// SkipMigrations gates the server's Postgres migrator at startup.
+	// Set true when migrations are applied out-of-band (e.g. by
+	// `arctl db migrate up` from CI/CD ahead of the rollout).
+	// Populated from AGENT_REGISTRY_SKIP_MIGRATIONS or the unprefixed
+	// SKIP_MIGRATIONS as a fallback (prefixed wins; see NewConfig).
```
iplay88keys: What's the benefit of having a prefix vs non-prefix here?

pkg/cli/db/migrate/migrate.go
```diff
@@ -0,0 +1,437 @@
...
+	cmd.AddCommand(newVersionCmd())
+	cmd.AddCommand(newGotoCmd())
+	cmd.AddCommand(newForceCmd())
+	return cmd
+}
+
+// EmbeddingsEnabledOverride returns (value, true) when the user passed
+// --embeddings-enabled on the command line and (false, false) otherwise.
+// Source-registration BuildConfig closures use the (set==false) signal
+// to fall through to the env var so the CLI's view of the embeddings
```
iplay88keys: Is the expectation that the cli is run on the same host/with the same env vars that agentregistry is run so that these env vars match?

pkg/cli/db/migrate/migrate.go
```diff
@@ -0,0 +1,437 @@
...
+					m := database.NewMigrator(conn, srcs[i].cfg)
+					applied, _, err := m.Status(ctx)
+					if err != nil {
+						return err
+					}
+					toRollback := min(remaining, len(applied))
+					if toRollback == 0 {
+						continue
+					}
+					if err := m.Down(ctx, toRollback); err != nil {
```
iplay88keys: As this is a CLI, should there be any form of user verification before the actual migration (either a `--dry-run` or `--non-interactive` mode)? It could be useful knowing what's going to change before the migration runs.

pkg/cli/db/migrate/migrate.go
iplay88keys: I think it would be useful to see the actual migration numbers here instead of just the number applied. Should this command just be merged with the `version` command?

pkg/cli/db/migrate/migrate.go
```diff
@@ -0,0 +1,437 @@
...
+				infos := make([]srcInfo, len(srcs))
+				targetIdx := -1
+				for i, src := range srcs {
+					mig := database.NewMigrator(conn, src.cfg)
+					low, high := sourceBoundsFromConfig(src.cfg)
+					infos[i] = srcInfo{m: mig, low: low, high: high}
+					// First-match wins (sources are pre-sorted by offset,
+					// so this is the lowest-offset match). Empty sources
+					// report a sentinel `(low, low-1)` range which matches
+					// no version, so they're naturally skipped here.
```
iplay88keys: Should there be an error if there's an overlap of ranges? Maybe that's already covered elsewhere, though.

pkg/registry/database/migrate.go
```diff
@@ -123,6 +158,11 @@ func (m *Migrator) loadMigrations() ([]Migration, error) {
 		if entry.IsDir() || !strings.HasSuffix(entry.Name(), ".sql") {
 			continue
 		}
+		// Skip .down.sql siblings in the main scan; they're loaded
+		// alongside their up counterpart below.
+		if strings.HasSuffix(entry.Name(), ".down.sql") {
+			continue
+		}
```
iplay88keys: We can get around this by always having `.down.sql` migrations (even retroactively added).

iplay88keys (commented):
Looks good overall. I think the biggest thing is that this requires significant custom migration code due to the custom migrator that exists here. This introduces additional risk of bugs that could be mitigated by using a standard migration library. If the codebase were to be switched to `golang-migrate`, for example, much of the complexity of this PR could be reduced. With the migration split, the custom offset logic can go away and make a path to moving toward using a standardized migration library.

internal/registry/config/config.go
nikolasmatt: We have dropped the prefix for env vars downstream but we haven't in OSS (i assume yet) this creates parity for `SKIP_MIGRATIONS` but doesn't break the old `AGENT_REGISTRY_SKIP_MIGRATIONS`

pkg/cli/db/migrate/migrate.go
nikolasmatt: This isn't natively supported by golang-migrate — we'd need to build custom code (either a file-listing dry-run, a SQL-body preview, or a tx-rollback executor that has its own caveats around non-transactional DDL like CREATE INDEX CONCURRENTLY). Worth tackling in a follow-up PR with its own design discussion.

pkg/cli/root.go
nikolasmatt: addressing this actually caught a real bug 👍

internal/registry/registry_app.go
```diff
@@ -74,7 +74,12 @@ func App(ctx context.Context, opts ...types.AppOptions) error {
 	}
 	authz := auth.Authorizer{Authz: authzProvider}
 
-	db, err := openDatabase(ctx, dbCtx, cfg, options, authz)
+	// Effective SkipMigrations: AppOptions wins when set, otherwise the
+	// env-driven Config value (AGENT_REGISTRY_SKIP_MIGRATIONS) applies.
+	// Either may be true; both being false means run the migrator.
+	skipMigrations := options.SkipMigrations || cfg.SkipMigrations
```
iplay88keys, line 81: This makes me wonder if it would be better to combine these ways of configuring AgentRegistry. i.e. read both AppOptions and env vars into a single struct with precedence one way or another.

nikolasmatt: I agree, but i'd rather tackle this separately. the scope of this PR is already large

iplay88keys: I completely agree 👍

pkg/registry/v1alpha1store/migrations/README.md
```diff
@@ -0,0 +1,112 @@
...
+the server (at startup, unless `SkipMigrations` is set) and the
+`arctl db migrate` CLI. The migrator is `golang-migrate/migrate v4`
+with the `pgx/v5` driver and the `iofs` source — see
+`pkg/registry/database/migrate.go`.
+
+## File-naming convention
+
+```
+NNN_short_name.up.sql        # forward (required)
+NNN_short_name.down.sql      # reverse (required by the iofs source)
```
iplay88keys, line 55: Can we check how this is stored via golang-migrate? It seems worth while to have `NNNNNN_` prefixed files so that we don't run out as quickly. If you can just expand that in golang-migrate, no worries, we can rename in the future. Otherwise, I think we should use this style off the bat.

iplay88keys (approved):
Looks good to me overall. As long as we have tested the following, I'm good:
- Fresh install passes e2e tests
- Upgrade from old version to new version still passes e2e tests

This will leave around the legacy tables, so we need to make sure to surface to users that they can clean those up at their discretion.

pkg/registry/v1alpha1store/migrations/README.md
nikolasmatt: these are stored in a `BigInt` column in the DB. the leading 0s only matter in the file name so we can easily address this when we get close to 999 migrations.

iplay88keys (approved):
LGTM

PR #505, opened 2026-05-19
Remove semantic search + embeddings

inFocus7:
note: [original pr](https://github.com/agentregistry-dev/agentregistry/pull/500) auto-closed since the underlying base pr merged

# Description

Removes the unwired semantic-search / embeddings feature. The HTTP handlers, the indexer, and the public Store surface are deleted; the `semantic_embedding*` columns and HNSW indexes are dropped by a new v1alpha1 migration on upgrade.

What's removed:
- `internal/registry/api/handlers/v0/embeddings/`: the indexer kickoff and job-status endpoints. Never wired into the running router.
- `internal/registry/embeddings/`: the indexer, helpers, and Provider interface. Never constructed at boot.
- ⚠️ `pkg/semantic/` and the `v1alpha1store.Store` semantic methods (`SetEmbedding`, `GetEmbeddingMetadata`, `SemanticList`, `VectorLiteral`). **Public-API breaking**: downstream consumers importing these symbols will get compile errors.
- `EmbeddingsConfig` and the `AGENT_REGISTRY_EMBEDDINGS_ENABLED` env var.
- The migrator's generic `MigratorConfig.Skip` predicate, which only ever gated 003.

Migrations:
- `003_embeddings.sql` is retained as a no-op (comment-only) for historical consistency with installs that have version 203 recorded in `schema_migrations`. Plain postgres images can apply it.
- `008_drop_semantic_embeddings.sql` drops the columns and HNSW indexes with `IF EXISTS`, safe on installs that never ran 003 originally.

# Change Type

```
/kind breaking_change
/kind cleanup
```

# Changelog

```release-note
Remove embeddings and semantic search.
⚠️ If you had previously enabled embeddings and use the pgvector postgres image, keep that image in place during the upgrade migration. Once the upgrade completes, you can `DROP EXTENSION vector;` and switch to a standard Postgres image post-release.
```

# Additional Notes

## Upgrade notes

Operators currently running with `AGENT_REGISTRY_EMBEDDINGS_ENABLED=true` on a postgres image that includes pgvector (e.g. `pgvector/pgvector:pg16`) must keep the pgvector image during this upgrade.

Migration 208 drops the semantic_embedding* columns and HNSW indexes, which are typed against pgvector. PostgreSQL requires the extension's shared library to be loadable on disk to execute the drops. Attempting to upgrade the application and switch to a plain postgres image in _the same_ Helm release fails with:

```
ERROR: could not access file "$libdir/vector": No such file or directory
```

To remove the pgvector dependency:
1. Upgrade the application first, keeping the pgvector image.
2. Verify migration 208 applied: SELECT version FROM schema_migrations WHERE version=208;
3. In a follow-up Helm release, swap the bundled postgres image to plain postgres of the same major version.

Operators who ran with EMBEDDINGS_ENABLED=false (the default) have nothing to drop and can upgrade freely on plain postgres.


nikolasmatt (commented):
Clean removal — no stale references in Go, charts, or docs; handler was never wired, confirmed. One important on 008, two nits.

Also: PR body's breaking-change list omits the `MigratorConfig` signature change (`MigratorConfig(bool)` → `MigratorConfig()`) and the removal of the exported `MigratorConfig.Skip` field. Downstream consumers will hit compile errors — worth listing alongside `pkg/semantic`.

pkg/registry/v1alpha1store/migrations/008_drop_semantic_embeddings.sql
```diff
@@ -0,0 +1,47 @@
...
+
+ALTER TABLE v1alpha1.prompts
+    DROP COLUMN IF EXISTS semantic_embedding,
+    DROP COLUMN IF EXISTS semantic_embedding_provider,
+    DROP COLUMN IF EXISTS semantic_embedding_model,
+    DROP COLUMN IF EXISTS semantic_embedding_dimensions,
+    DROP COLUMN IF EXISTS semantic_embedding_checksum,
+    DROP COLUMN IF EXISTS semantic_embedding_generated_at;
+
+DROP EXTENSION IF EXISTS vector;
```
nikolasmatt: **important** — `DROP EXTENSION` without `CASCADE` raises `2BP01` if anything else in the DB depends on `vector` (operator's own use, future downstream extension with `VersionOffset=500+`). The migrator wraps the file in a single txn (`pkg/registry/database/migrate.go:212`), so failure rolls back the column drops too and `schema_migrations` never records v208 — next boot retries the same failure.

Dropping the columns + indexes alone fully severs agentregistry from pgvector; the extension drop is cosmetic. Recommend deleting this line and documenting that operators can `DROP EXTENSION vector` manually.

pkg/registry/v1alpha1store/migrations/003_embeddings.sql
```diff
@@ -1,84 +1,21 @@
...
-CREATE INDEX IF NOT EXISTS v1alpha1_prompts_semantic_embedding_hnsw
-    ON v1alpha1.prompts USING hnsw (semantic_embedding vector_cosine_ops);
-
--- Providers + Deployments don't participate in semantic search today (users
--- search Agents/MCPServers/Skills/Prompts for capabilities; Providers are
--- infrastructure metadata and Deployments are lifecycle state). Columns
--- stay off these tables to keep the schema honest — add them here if the
--- search model changes.
+-- Comment-only SQL is valid input: PostgreSQL parses zero statements
+-- and the migrator records the version as applied.
```
nikolasmatt: **nit** — comment-only relies on pgx accepting `EmptyQueryResponse`. True today, but a trailing `SELECT 1;` would harden against strict-mode tooling and signal intent to a reader scanning the migrations dir.

scripts/kind/README.md
```diff
@@ -55,16 +53,6 @@ Then connect with psql:
...
-
-The `pgvector` extension is automatically available via the `pgvector/pgvector` image. The AgentRegistry server enables it on startup.
-
-To verify manually:
-
-```sql
-SELECT * FROM pg_extension WHERE extname = 'vector';
-```
-
 ### Data Persistence
```
nikolasmatt, line 56: **nit** (not about this line — about `charts/agentregistry/tests/postgresql_test.yaml:52`, which isn't in this PR's diff): the test `it: Deployment image can be overridden to pgvector via component fields` now reads like pgvector is still a first-class capability. Consider renaming to `… overridden to a custom image via component fields` with pgvector as just the example value.

nikolasmatt (approved):

PR #511, opened 2026-05-22
Resource name validation (DNS-1123 subdomain) + decouple MCP Package Server name from resource name

inFocus7:
# Description

We currently have a few issues in our system:
1. Throughout our codebase we flip-flop on requiring resource name sto follow `PREFIX/NAME` convention - iirc ui + cli both differ in requirements.
2. We mistakenly split `NAMESPACE/NAME` on a resource name into the {metadata.namespace, metadata.name} form - while we currently haven't fully implemented the differing namespace use-case.
    - When defining the resource name, we should not parse this - we should treat is as the name only.
3. For MCP validation, we compare the resource name against the package name in the upstream registry
    - This is currently _not_ an issue, but will be if we tighten the name validation, since mcp registry naming convention is `REVERSE_DNS/NAME`, unrelated to our namespaces.

To resolve the above, the TL;DR:
1. We now have **strict** DNS-1123 subdomain compliancy requirement for resource names - similar to k8s resource name https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names. 
    - This fixes the UX to follow this new naming convention.
2. We added a new field `serverName` for MCP packages from external registries to use during validation.

Changes:
- Added new field for MCP Packages (used during package validation): `serverName`
    - [pkg/api/v1alpha1/mcpserver.go](https://github.com/agentregistry-dev/agentregistry/compare/main...inFocus7:agentregistry:infocus7/fix-mcp-server-naming-ux?expand=1#diff-2a2fecb82623b56416959f14b62311cdc20e7982f7cb083005bef5e45419b314)
- Added validation conditions for the new naming:`metadata.name` -> DNS-label on all resources + `[...].package.serverName` -> `PREFIX/NAME`
    - [pkg/api/v1alpha1/validation.go](https://github.com/agentregistry-dev/agentregistry/compare/main...inFocus7:agentregistry:infocus7/fix-mcp-server-naming-ux?expand=1#diff-1698712b40b833508bf6ce8c2bcdd471bf335af362354238f8218209387207a1), [pkg/api/v1alpha1/mcpserver_validate.go](https://github.com/agentregistry-dev/agentregistry/compare/main...inFocus7:agentregistry:infocus7/fix-mcp-server-naming-ux?expand=1#diff-ece2c1b8663de70f6a09e91de4db921cbb3564e558ac967190e3b6787e8908bd), [pkg/api/v1alpha1/registry_validate.go](https://github.com/agentregistry-dev/agentregistry/compare/main...inFocus7:agentregistry:infocus7/fix-mcp-server-naming-ux?expand=1#diff-f9b2914ac71abf287ab789673b479cbe132e44401a2343685238140679e14674)
- Updated MCP CLI `init` command to client-side validate based on new DNS-label requirements
    - [internal/cli/declarative/init.go](https://github.com/agentregistry-dev/agentregistry/compare/main...inFocus7:agentregistry:infocus7/fix-mcp-server-naming-ux?expand=1#diff-ebf172bcd400568908694626f99051884ea4e986611cdf718a0c69de96d16490)
- Updated UI component to validate based on new DNS-label requirement + added new `mcpPackage` field for MCP.
    - [ui/components/add-server-dialog.tsx](https://github.com/agentregistry-dev/agentregistry/compare/main...inFocus7:agentregistry:infocus7/fix-mcp-server-naming-ux?expand=1#diff-ab15e6c7e51fb09ffb931bf79f54712a0ee43995f08622590512b5964416d42d)
- Updated MCP UI server shim to **not** parse into "`NAMESPACE/NAME`", and instead do `{namespace: "default", name: NAME}`
    - [ui/lib/ui-shims.ts](https://github.com/agentregistry-dev/agentregistry/compare/main...inFocus7:agentregistry:infocus7/fix-mcp-server-naming-ux?expand=1#diff-1ef4f4938201fa55e1e770cf735dcb01069507a4caeadf7234fc5914f352ddde)
- Updated tests to expect & use correct DNS-label naming for resources
    - [e2e/declarative_test.go](https://github.com/agentregistry-dev/agentregistry/compare/main...inFocus7:agentregistry:infocus7/fix-mcp-server-naming-ux?expand=1#diff-3c1512cab4f30752e1f9970856bb203a57e7d922bae43f6f1f877109ea12f838), [e2e/init_build_run_test.go](https://github.com/agentregistry-dev/agentregistry/compare/main...inFocus7:agentregistry:infocus7/fix-mcp-server-naming-ux?expand=1#diff-ea453ce89ab2630303ae9e175c99c1ca1c6b23dfed478d8b3c46266bea1e9054), [internal/cli/declarative/init_test.go](https://github.com/agentregistry-dev/agentregistry/compare/main...inFocus7:agentregistry:infocus7/fix-mcp-server-naming-ux?expand=1#diff-3115dc6b2f49130451912b11843f5e71702c7cc9509d5a8ed384e80a96def116), [internal/cli/mcp/manifest/manager_envelope_test.go](https://github.com/agentregistry-dev/agentregistry/compare/main...inFocus7:agentregistry:infocus7/fix-mcp-server-naming-ux?expand=1#diff-8dfb26438d17465a66465f2b523119d6149cec91259c7ba4a82216a357c2974f), [internal/cli/scheme/scheme_test.go](https://github.com/agentregistry-dev/agentregistry/compare/main...inFocus7:agentregistry:infocus7/fix-mcp-server-naming-ux?expand=1#diff-752c374f9978267c71a43768f589eeb57540e2522b2936be4c00e42f981483b2), [pkg/api/v1alpha1/validation_test.go](https://github.com/agentregistry-dev/agentregistry/compare/main...inFocus7:agentregistry:infocus7/fix-mcp-server-naming-ux?expand=1#diff-6d3de1a2a198aa970520f8b84c1d69a836c7379a97d1ea44f5f399f0c25e0330), [ui/components/__tests__/server-card.test.tsx](https://github.com/agentregistry-dev/agentregistry/compare/main...inFocus7:agentregistry:infocus7/fix-mcp-server-naming-ux?expand=1#diff-07a870b52b51dd1301ef956ad1944dab05db1da2639ac517ec99e65220473afc)
- Updated examples to use the new naming
    - [examples/agent.yaml](https://github.com/agentregistry-dev/agentregistry/compare/main...inFocus7:agentregistry:infocus7/fix-mcp-server-naming-ux?expand=1#diff-ad808fab9a63386794a5328dec23ad9b09f673dc590b1a3e1b356806d27f3a66), [examples/full-stack.yaml](https://github.com/agentregistry-dev/agentregistry/compare/main...inFocus7:agentregistry:infocus7/fix-mcp-server-naming-ux?expand=1#diff-a6d3d1c1314b2221a96b8420bf438897df81d796b2e64d673a51e00a2a2ce38e), [examples/mcp-remote.yaml](https://github.com/agentregistry-dev/agentregistry/compare/main...inFocus7:agentregistry:infocus7/fix-mcp-server-naming-ux?expand=1#diff-625bd8d405d9e24caaaab37790e5989e92f388b9b9e98a4002f920191bb3aab4), [examples/mcp.yaml](https://github.com/agentregistry-dev/agentregistry/compare/main...inFocus7:agentregistry:infocus7/fix-mcp-server-naming-ux?expand=1#diff-030c1283cdd5ebd41ddd1d48942edd36d262f95d686caa62698e11f88c6aa55b), 
- Updated CLI to update examples using `NAME` DNS-label compliancy
    - [internal/cli/declarative/delete.go](https://github.com/agentregistry-dev/agentregistry/compare/main...inFocus7:agentregistry:infocus7/fix-mcp-server-naming-ux?expand=1#diff-cb6c1da4a7d8a02b05ce5702629b4f97e1a212cb4ffbd498a6074e8ea4778694), [internal/cli/declarative/init.go](https://github.com/agentregistry-dev/agentregistry/compare/main...inFocus7:agentregistry:infocus7/fix-mcp-server-naming-ux?expand=1#diff-ebf172bcd400568908694626f99051884ea4e986611cdf718a0c69de96d16490)

> **NOTE**: These are **breaking**. There is more info on it in a below section.

# Change Type

```
/kind breaking_change
/kind fix
```

# Changelog

```release-note
AgentRegistry resource names must now use DNS-1123 subdomain format (lowercase letters, numbers, hyphens, and periods). MCP package manifests now use `spec.source.package.serverName` for the upstream MCP server name, so manifests that used slash-delimited resource names such as `namespace/name` must rename those resources and set `serverName` when package validation needs the upstream name.
```

# Additional Notes

## Breaking Changes

Due to the new constraints, this is a breaking change to **all** resources & references to them. Users must update their declarative YAMLs to follow this new naming requirement.

## Local Validation (Script)

_thanks to Claude for this local setup script to manually validate CLI, you can also validate UI on the running instance_

```sh
#!/usr/bin/env bash
#
# Interactive validation harness for the DNS-1123 subdomain naming rule across
# v1alpha1 resource kinds (Agent, Skill, Prompt, Deployment, MCPServer).
#
# 1. Builds + starts the daemon against the currently checked-out branch.
# 2. Probes the validator with several non-compliant names across kinds and
#    asserts each one is rejected with a DNS-1123 error.
# 3. Applies a small set of compliant fixtures (MCPServer + Agent + Skill +
#    Prompt) and lists them.
# 4. Pauses so you can poke the UI / CLI yourself before cleanup.
#
# Trap-on-exit purges the daemon so the next run starts from a clean DB.
#
# Requires: clean tracked-file working tree (untracked OK), Docker daemon
# running, current directory is the repo root, AND network access to the npm
# registry (the package-shape MCPServer fixture in step 2 exercises the npm
# ownership validator against @modelcontextprotocol/server-memory).

set -euo pipefail

API_URL="http://localhost:12121/v0"
UI_URL="http://localhost:12121"
ARCTL="./bin/arctl"
TMP_DIR="$(mktemp -d -t dns-validate.XXXXXX)"
trap 'rm -rf "$TMP_DIR"' EXIT

# ----- preflight -----------------------------------------------------------

if ! git diff --quiet || ! git diff --cached --quiet; then
    echo "Tracked files have uncommitted changes. Commit or stash before running." >&2
    echo "(Untracked files are ignored — only modifications to tracked files block.)" >&2
    git status --short >&2
    exit 1
fi

# ----- cleanup trap (purge daemon) -----------------------------------------

cleanup() {
    local code=$?
    echo
    echo "=== cleanup: purging daemon ==="
    make daemon-stop-purge >/dev/null 2>&1 || true
    exit "$code"
}
trap cleanup EXIT INT TERM

# ----- helpers -------------------------------------------------------------

daemon_health() { curl -sf "$API_URL/health" >/dev/null 2>&1; }

wait_for_daemon() {
    for _ in $(seq 1 60); do
        if daemon_health; then return 0; fi
        sleep 1
    done
    return 1
}

apply_stdin() {
    "$ARCTL" apply -f - --registry-url "$API_URL"
}

# probe_rejection feeds a single-document YAML payload via stdin and asserts
# that the apply fails with an error matching the given pattern. Returns 0 on
# correct rejection, 1 on unexpected acceptance or wrong error shape.
probe_rejection() {
    local label="$1"
    local pattern="$2"
    local payload="$3"
    local out
    out=$(printf '%s' "$payload" | "$ARCTL" apply -f - --registry-url "$API_URL" 2>&1 || true)
    if echo "$out" | grep -qiE "$pattern"; then
        echo "  ✓ rejected: $label"
        return 0
    fi
    echo "  ✗ FAIL: $label was NOT rejected with the expected error"
    echo "    expected pattern: $pattern"
    echo "    output: $out"
    return 1
}

# probe_acceptance feeds a single-document YAML payload via stdin and asserts
# the apply succeeds. Mirrors probe_rejection for symmetric reporting.
probe_acceptance() {
    local label="$1"
    local payload="$2"
    local out
    out=$(printf '%s' "$payload" | "$ARCTL" apply -f - --registry-url "$API_URL" 2>&1 || true)
    # arctl apply uses exit code via "✗ ... failed:" line on per-doc failure;
    # accept either a non-error exit or a stdout containing "✓".
    if echo "$out" | grep -qE "✓|configured|created|unchanged"; then
        echo "  ✓ accepted: $label"
        return 0
    fi
    echo "  ✗ FAIL: $label was REJECTED but should have been accepted"
    echo "    output: $out"
    return 1
}

# ----- run -----------------------------------------------------------------

CURRENT_BRANCH="$(git rev-parse --abbrev-ref HEAD 2>/dev/null || echo '<unknown>')"
echo "Validation harness for DNS-1123 subdomain naming rule"
echo "  branch under test: $CURRENT_BRANCH"
echo "  daemon URL:        $API_URL"
echo

echo "→ purge any prior daemon state"
make daemon-stop-purge >/dev/null 2>&1 || true

echo "→ rebuild CLI + server image for $CURRENT_BRANCH"
make build-cli docker docker-tag-as-dev >/dev/null

echo "→ start daemon"
make daemon-start
wait_for_daemon || { echo "Daemon failed to come up." >&2; exit 1; }

# --- step 1: negative-path probes across kinds ---

echo
echo "→ verifying non-DNS-subdomain names are rejected across kinds"
neg_failures=0

DNS_PATTERN="DNS-1123 subdomain|invalid format"
OWNERSHIP_PATTERN="ownership validation failed|Expected mcpName"

# MCPServer — slash
probe_rejection "MCPServer name 'acme/foo' (slash)" "$DNS_PATTERN" "$(cat <<'EOF'
apiVersion: ar.dev/v1alpha1
kind: MCPServer
metadata:
  name: acme/foo
spec:
  remote: { type: streamable-http, url: https://mcp.example.com/x }
EOF
)" || neg_failures=$((neg_failures + 1))

# MCPServer — uppercase
probe_rejection "MCPServer name 'MyServer' (uppercase)" "$DNS_PATTERN" "$(cat <<'EOF'
apiVersion: ar.dev/v1alpha1
kind: MCPServer
metadata:
  name: MyServer
spec:
  remote: { type: streamable-http, url: https://mcp.example.com/x }
EOF
)" || neg_failures=$((neg_failures + 1))

# MCPServer — > 63 chars
long_name=$(printf 'a%.0s' {1..64})
probe_rejection "MCPServer name (>63 chars)" "$DNS_PATTERN" "$(cat <<EOF
apiVersion: ar.dev/v1alpha1
kind: MCPServer
metadata:
  name: $long_name
spec:
  remote: { type: streamable-http, url: https://mcp.example.com/x }
EOF
)" || neg_failures=$((neg_failures + 1))

# Skill — underscore
probe_rejection "Skill name 'my_skill' (underscore)" "$DNS_PATTERN" "$(cat <<'EOF'
apiVersion: ar.dev/v1alpha1
kind: Skill
metadata:
  name: my_skill
spec:
  description: skill with bad name
  source:
    repository: { url: https://github.com/example/repo }
EOF
)" || neg_failures=$((neg_failures + 1))

# Prompt — leading dot (still rejected — dots themselves are allowed by
# DNS-1123 subdomain, but leading/trailing/double dots are not)
probe_rejection "Prompt name '.my-prompt' (leading dot)" "$DNS_PATTERN" "$(cat <<'EOF'
apiVersion: ar.dev/v1alpha1
kind: Prompt
metadata:
  name: .my-prompt
spec:
  content: hello
EOF
)" || neg_failures=$((neg_failures + 1))

# Agent ref — bad ref name in spec.mcpServers
probe_rejection "Agent ref spec.mcpServers[].name 'acme/foo' (slash)" "$DNS_PATTERN" "$(cat <<'EOF'
apiVersion: ar.dev/v1alpha1
kind: Agent
metadata:
  name: valid-agent
spec:
  image: ghcr.io/example/agent:latest
  description: agent referencing a slash-named MCPServer
  language: python
  framework: adk
  modelProvider: gemini
  modelName: gemini-2.0-flash
  mcpServers:
    - kind: MCPServer
      name: acme/foo
      tag: latest
EOF
)" || neg_failures=$((neg_failures + 1))

# Package ownership — DNS-subdomain name, but serverName doesn't match what
# @modelcontextprotocol/server-memory publishes as its mcpName upstream. The
# npm ownership validator must reject this with an ownership-mismatch error,
# proving the serverName ↔ upstream cross-check actually fires.
probe_rejection "MCPServer package serverName mismatch (npm ownership)" "$OWNERSHIP_PATTERN" "$(cat <<'EOF'
apiVersion: ar.dev/v1alpha1
kind: MCPServer
metadata:
  name: validate-bad-ownership
spec:
  title: Mismatched serverName, should be rejected by npm ownership validator
  source:
    package:
      registryType: npm
      identifier: "@modelcontextprotocol/server-memory"
      version: latest
      serverName: "io.github.example/wrong-name"
      transport:
        type: stdio
EOF
)" || neg_failures=$((neg_failures + 1))

if (( neg_failures > 0 )); then
    echo
    echo "Validator probe failed: $neg_failures payload(s) were NOT rejected as expected." >&2
    exit 1
fi

echo
echo "→ verifying compliant boundary cases are accepted"
pos_failures=0

# Single-char name — DNS-1123 subdomain accepts 1-char names.
probe_acceptance "MCPServer name 'a' (single char)" "$(cat <<'EOF'
apiVersion: ar.dev/v1alpha1
kind: MCPServer
metadata:
  name: a
spec:
  remote: { type: streamable-http, url: https://mcp.example.com/a-only }
EOF
)" || pos_failures=$((pos_failures + 1))

# Dotted reverse-DNS name — newly allowed by subdomain (was rejected under label).
probe_acceptance "MCPServer name 'io.example.mcp' (dotted reverse-DNS)" "$(cat <<'EOF'
apiVersion: ar.dev/v1alpha1
kind: MCPServer
metadata:
  name: io.example.mcp
spec:
  remote: { type: streamable-http, url: https://mcp.example.com/dotted }
EOF
)" || pos_failures=$((pos_failures + 1))

# Hyphen + dot in same name — common ecosystem shape (e.g. mcp-server.example.io).
probe_acceptance "Skill name 'data-tools.example.io' (hyphen + dot)" "$(cat <<'EOF'
apiVersion: ar.dev/v1alpha1
kind: Skill
metadata:
  name: data-tools.example.io
spec:
  description: skill with hyphen and dot
  source:
    repository: { url: https://github.com/example/repo }
EOF
)" || pos_failures=$((pos_failures + 1))

if (( pos_failures > 0 )); then
    echo
    echo "Validator probe failed: $pos_failures payload(s) were REJECTED but should have been accepted." >&2
    exit 1
fi

# --- step 2: apply compliant fixtures across kinds ---

echo
echo "→ applying compliant DNS-subdomain resources"
"$ARCTL" apply -f - --registry-url "$API_URL" <<'EOF'
apiVersion: ar.dev/v1alpha1
kind: MCPServer
metadata:
  name: validate-mcp-a
spec:
  title: Validation MCP A
  remote:
    type: streamable-http
    url: https://mcp.example.com/a
---
apiVersion: ar.dev/v1alpha1
kind: MCPServer
metadata:
  name: validate-mcp-b
spec:
  title: Validation MCP B
  remote:
    type: streamable-http
    url: https://mcp.example.com/b
---
# Package-shape MCPServer with the new spec.source.package.serverName field.
# Exercises the upstream ownership validator (npm metadata lookup) AND the UI's
# new "Server Name" sidebar row. Requires network access to the npm registry
# AND that `@modelcontextprotocol/server-memory`'s published package carries
# `mcpName: io.github.modelcontextprotocol/server-memory` — if either drifts,
# the apply will fail with an ownership-validation error.
apiVersion: ar.dev/v1alpha1
kind: MCPServer
metadata:
  name: validate-mcp-memory
spec:
  title: Memory MCP (npm package, ownership-validated)
  source:
    package:
      registryType: npm
      identifier: "@modelcontextprotocol/server-memory"
      version: latest
      serverName: "io.github.modelcontextprotocol/server-memory"
      transport:
        type: stdio
---
apiVersion: ar.dev/v1alpha1
kind: Skill
metadata:
  name: validate-skill
spec:
  description: validation harness skill
  source:
    repository:
      url: https://github.com/example/repo
---
apiVersion: ar.dev/v1alpha1
kind: Prompt
metadata:
  name: validate-prompt
spec:
  content: You are a validation prompt.
---
apiVersion: ar.dev/v1alpha1
kind: Agent
metadata:
  name: validate-agent
spec:
  image: ghcr.io/example/agent:latest
  description: References two compliant remote MCPServers.
  language: python
  framework: adk
  modelProvider: gemini
  modelName: gemini-2.0-flash
  mcpServers:
    - kind: MCPServer
      name: validate-mcp-a
      tag: latest
    - kind: MCPServer
      name: validate-mcp-b
      tag: latest
EOF

# --- step 3: pause for manual inspection ---

echo
echo "────────────────────────────────────────────────────────────────"
echo "   UI:   $UI_URL"
echo "   API:  $API_URL"
echo "   CLI:  $ARCTL get mcp     --registry-url $API_URL"
echo "         $ARCTL get agent   --registry-url $API_URL"
echo "         $ARCTL get skill   --registry-url $API_URL"
echo "         $ARCTL get prompt  --registry-url $API_URL"
echo "────────────────────────────────────────────────────────────────"
echo "Press any key to finish (will purge daemon on exit)."
echo "(Ctrl-C also works; trap will purge.)"
read -rsn1
echo
echo "Done."
```

## Screenshots

ui validation


pkg/api/v1alpha1/mcpserver.go
```diff
@@ -61,10 +61,27 @@ type MCPTransport struct {
...
+	//
+	// When set, the registry's ownership validator compares this against
+	// the package's published claim. When unset, the validator compares
+	// against metadata.name.
+	//
+	// Set this when `metadata.name` (DNS-1123 label compliant) cannot
+	// represent an upstream registry's naming shape (e.g. `domain/name`) shape.
+	//
+	// Optional.
+	MCPName string `json:"mcpName,omitempty" yaml:"mcpName,omitempty"`
```
shashankram: The Reverse DNS name convention applies to server names, so the field should be indicative of that
```suggestion
	ServerName string `json:"serverName,omitempty" yaml:"serverName,omitempty"`
```

pkg/registry/v1alpha1store/migrations/009_mcp_dns_label.sql
```diff
@@ -0,0 +1,204 @@
+-- Tighten MCPServer.metadata.name to DNS-1123 label (matches the application
```
shashankram: I am against adding a DB migration file for this. Either we allow metadata.Name to be backward compatible or treat this as a fully breaking name change.

pkg/api/v1alpha1/mcpserver.go
inFocus7: I'm leaning towards `mcpName` being the better name for this field. It mirrors the naming on the upstream packages, which is `mcpName` + `mcp-name` + the doc comment states the purpose (i could update the doc comment if it's not clear enough). This would make it easier for users adding packages to compare and easily see what the field refers to

https://raw.githubusercontent.com/modelcontextprotocol/servers/refs/heads/main/src/fetch/README.md
```
<!-- mcp-name: io.github.modelcontextprotocol/server-fetch -->
```

https://raw.githubusercontent.com/modelcontextprotocol/servers/refs/heads/main/src/everything/package.json
```
  "mcpName": "io.github.modelcontextprotocol/server-everything",
```

pkg/registry/v1alpha1store/migrations/009_mcp_dns_label.sql
inFocus7: I'm fine with either. I added this file thinking with DNS label compliancy being a breaking change, but added migration because AFAICT we don't allow editing MCP server names. Without the migration file, a user would need to recreate their mcp server resources.

---

For backwards compatibility, this PR would be ~80% slimmer. We would:
1. remove migration file + test
2. remove `/` _requirement_, but allow it
3. likely keep tests as-is, as they should work

I think/hope we already handle percent-encoding for `/`, and if so, it should be fast to handle

pkg/api/v1alpha1/mcpserver.go
shashankram: MCPPackage.MCPName is odd. Even https://modelcontextprotocol.io/registry/about#verifying-server-authenticity specifies that the reverse DNS constraint applies to the "server name"

pkg/registry/v1alpha1store/migrations/009_mcp_dns_label.sql
inFocus7: before closing this, going forward with breaking change here + other resources

docs/declarative-cli.md
```diff
@@ -10,7 +10,7 @@ arctl build summarizer/ --push    # optional: build and push Docker image
 arctl apply -f summarizer/agent.yaml
 ```
 
-`arctl init agent NAME` and `arctl init mcp NAMESPACE/NAME` pick a framework + language interactively unless `--framework` and `--language` are provided. Run `arctl init agent NAME` (or `arctl init mcp NAMESPACE/NAME`) on its own to see the available choices.
+`arctl init agent NAME` and `arctl init mcp NAME` pick a framework + language interactively unless `--framework` and `--language` are provided. Run `arctl init agent NAME` (or `arctl init mcp NAME`) on its own to see the available choices. All resource `metadata.name` values (Agent, Skill, Prompt, Deployment, MCPServer) must be DNS-1123 label: lowercase alphanumeric and hyphens only, max 63 chars, must start and end with alphanumeric. Agent names additionally cannot collide with Python keywords (`class`, `import`, `return`, …) because they become Python identifiers in generated code.
```
shashankram: we don't need to limit these to 63 chars. CR's metadata.name in k8s have a 253 limit, so I'd have that limit here instead.

inFocus7: updated to 253, also now following dns subdomain rules (lowercase, alphanum, `-`, `.`)

docs/declarative-cli.md
```diff
@@ -10,7 +10,7 @@ arctl build summarizer/ --push    # optional: build and push Docker image
 arctl apply -f summarizer/agent.yaml
 ```
 
-`arctl init agent NAME` and `arctl init mcp NAMESPACE/NAME` pick a framework + language interactively unless `--framework` and `--language` are provided. Run `arctl init agent NAME` (or `arctl init mcp NAMESPACE/NAME`) on its own to see the available choices.
+`arctl init agent NAME` and `arctl init mcp NAME` pick a framework + language interactively unless `--framework` and `--language` are provided. Run `arctl init agent NAME` (or `arctl init mcp NAME`) on its own to see the available choices. All resource `metadata.name` values (Agent, Skill, Prompt, Deployment, MCPServer) must be DNS-1123 subdomain: lowercase alphanumeric, hyphens, and dots; max 253 chars; each dot-separated segment must start and end with alphanumeric (max 63 chars per segment). Examples: `my-server`, `io.example.mcp`. Agent names additionally cannot contain hyphens or dots and cannot collide with Python keywords (`class`, `import`, `return`, …) because they become Python identifiers in generated code.
```
shashankram, line 13: > Agent names additionally cannot contain hyphens or dots and cannot collide with Python keywords (`class`, `import`, `return`, …) because they become Python identifiers in generated code.

This is a weird requirement as resource names should not be conflated with python import names. What is the exact issue here? Can we not sanitize metadata.name when using with Python imports/identifiers?

inFocus7: Im not entirely sure, this was pre-existing code that I documented (and moved a bit) alongside the new DNS-subdomain name requirements. I can look into this, but this is pre-existing behavior

I believe this is primarily an limitation with initializing an agent, although publishing an agent resource into the registry allows for the DNS subdomain naming

shashankram: Can we fix this and relax this requirement? Otherwise we have inconsistent resource naming requirements across Kinds and that's undesirable

inFocus7: I think this should be a follow-up to not block this breaking change from merging

The agent name is used throughout different portions of the agent code + templating (e.g. directory names, python identifiers, declarative agent yaml, Dockerfiles,) so there is risk of introducing bugs unrelated to this work

internal/cli/declarative/delete.go
```diff
@@ -40,7 +40,7 @@ TYPE must be one of: agent, mcp, skill, prompt, deployment
   arctl delete -f my-server/mcp.yaml
   arctl delete agent acme/summarizer --tag stable
   arctl delete agent acme/summarizer --all-tags
-  arctl delete mcp acme/fetch --tag stable
+  arctl delete mcp acme-fetch --tag stable
```
shashankram, line 43: Why are `/` allowed in agent Names above?

pkg/api/v1alpha1/registry_validate.go
```diff
@@ -67,20 +67,25 @@ func validatePackages(
...
+// entry (includes RegistryBaseURL + FileSHA256). The claim name passed
+// to the per-registry validator is p.ServerName when set, else m.Metadata.Name.
 func (m *MCPServer) ValidateRegistries(ctx context.Context, v RegistryValidatorFunc) error {
 	if v == nil || m.Spec.Source == nil || m.Spec.Source.Package == nil {
 		return nil
 	}
 	p := m.Spec.Source.Package
+	claimName := p.ServerName
+	if claimName == "" {
+		claimName = m.Metadata.Name
```
shashankram: What is claimName meant to do? Conflating this with metadata.name seems wrong so I question this fallback

inFocus7: Ownership "claim" name to check during validation. First it checks for the new `package.serverName` name, and if unset it falls back to the resource name. This value is what gets compared to the package's name.

We could also remove this, since it's very likely that the resource name wouldn't match the package one during validation (due to `/`), but I wanted to have a fallback to the original resource name

internal/cli/declarative/delete.go
inFocus7: Cleaned it up. There's a lot of stale references with `/` in docs & tests (that don't go through the validation path, so aren't affected by validation), but I believe I got them all now

pkg/api/v1alpha1/registry_validate.go
shashankram: I don't think the fallback logic makes sense here. If serverName is a required value for MCPPackage, why is it needed? If it's optional, then we should skip claim validation. 

I still don't get what you mean by "ownership claim name to check". What exactly is being validated here and against what?

inFocus7: It seems like it's primarily for trust, ensuring the fetched package matches what a user declares in their mcp server package definition.

It's optional because `metadata.name` can be used as a viable fallback value. This is pre-existing validation, and I'm not sure how much it matters. I was primarily focused on resolving the naming issues, and needed to maintain a way to define the name used during the validation step.

I don't think leaving it empty should be an _implicit_ skip, since it would be easy to miss this, and silently disabling a security check due to a missing field feels wrong.

shashankram: If it's a security check then serverName shouldn't be optional. `metadata.name` should not imply any security semantics in our system, it's purely a resource name

pkg/api/v1alpha1/mcpserver_validate.go
```diff
@@ -13,6 +13,23 @@ func (m *MCPServer) Validate() error {
...
+// the `name` field.
+func validateMCPPackageName(s string) error {
+	if s == "" {
+		return nil // optional field
+	}
+	if l := len(s); l < UpstreamMCPPackageNameMinLen || l > UpstreamMCPPackageNameMaxLen {
+		return fmt.Errorf("%w: serverName length must be %d-%d chars, got %d", ErrInvalidFormat, UpstreamMCPPackageNameMinLen, UpstreamMCPPackageNameMaxLen, l)
+	}
+	if !UpstreamMCPPackageNameRegex.MatchString(s) {
+		return fmt.Errorf("%w: serverName must match upstream pattern `namespace/name` (e.g. \"io.github.user/server\"): %q", ErrInvalidFormat, s)
```
shashankram, line 28:
```suggestion
		return fmt.Errorf("%w: got %s", ErrInvalidMCPServerNameFormat, s)
```
where
```
ErrInvalidMCPServerNameFormat = errors.New("invalid MCP serverName format. must be reverse DNS format like io.github.user/server"
```

pkg/api/v1alpha1/registry_validate.go
inFocus7: It's optional, because there is a type that does not do this validation check (MCPB).

i can look into doing type-based validation so that it's _conditionally required_ (for all, except MCPB package types).

if we make `serverName` required and no longer fallback, we'll need to relax its regex to allow strings without `/`. OCI labels can have single-segmented names (no `/`), so `serverName` would need to have lax validation (if any) to work.

shashankram: All I am saying is that using metadata.name (a resource name in our system) for a security check seems very odd. I'd rather have explicit behavior where serverName is always used if this is a security check.

inFocus7: 👍🏼 I'm working on making it a required field for non-MCPBs.

The main difference would be that the `serverName` regex check would need to be more lax to allow for more variety of naming (e.g. OCI image labels without `/`), but that not a huge deal

shashankram (approved):

inFocus7:
Transient failure in MQ. Retrying
```
There was an issue establishing a connection while requesting https://fonts.googleapis.com/css2?family=Plus+Jakarta+Sans:wght@200..800&display=swap
```

PR #517, opened 2026-05-27
Add deployment controller foundations

ilackarms:
# Description

Adds durable Deployment controller foundations for v1alpha1 resources.

Key pieces:
- registers v1alpha1 kinds explicitly and keeps kind policy in OSS;
- adds durable control-plane invalidation events and reconcile work stores;
- moves Deployment apply/delete side effects behind controller-owned work claiming;
- wires controller config, retention pruning, LISTEN wakeups, and reconnect behavior;
- updates deployment logs/delete/API/CLI paths for the controller-owned lifecycle;
- removes the legacy deployment delete force bypass.

# Change Type

/kind feature
/kind cleanup

# Changelog

```release-note
Adds durable Deployment controller foundations for v1alpha1 reconciliation.
```

# Additional Notes

Verification:
- `go test ./...`
- `go test -tags=integration ./...`
- `env -u NODE_ENV make verify`
- GitHub checks are green on the current PR head.

Enterprise companion PR will pin to this OSS commit before its own review/merge.


internal/registry/api/handlers/v0/deploymentlogs/deploymentlogs.go
```diff
@@ -13,21 +13,25 @@ import (
...
+// Config bundles the inputs for Register. The resolver drives
 // adapter.Logs; the store fetches the Deployment row so the endpoint
 // can reject 404s early.
 type Config struct {
-	BasePrefix  string
-	Store       *v1alpha1store.Store
-	Coordinator *deploymentsvc.Coordinator
+	BasePrefix string
+	Store      *v1alpha1store.Store
+	Resolver   LogResolver
```
shashankram:
```suggestion
	LogResolver   LogResolver
```

internal/registry/config/config.go
```diff
@@ -31,6 +32,15 @@ type Config struct {
...
 	Verbose    bool   `env:"VERBOSE" envDefault:"false"`
+
+	// Controller foundation retention. These settings bound the KRT
+	// controller's durable invalidation/work tables. A zero duration disables
+	// pruning for that table.
+	ControllerEventRetention           time.Duration `env:"CONTROLLER_EVENT_RETENTION" envDefault:"24h"`
+	ControllerEventKeepAfterRevision   int64         `env:"CONTROLLER_EVENT_KEEP_AFTER_REVISION" envDefault:"0"`
+	ControllerWorkRetention            time.Duration `env:"CONTROLLER_WORK_RETENTION" envDefault:"24h"`
+	ControllerAttemptRetention         time.Duration `env:"CONTROLLER_ATTEMPT_RETENTION" envDefault:"168h"`
+	ControllerRetentionPruneBatchLimit int           `env:"CONTROLLER_RETENTION_PRUNE_BATCH_LIMIT" envDefault:"500"`
```
shashankram: Each of these need comments describing the setting

internal/registry/controller/deployment.go
```diff
@@ -0,0 +1,89 @@
...
+
+	"github.com/agentregistry-dev/agentregistry/pkg/api/v1alpha1"
+	"github.com/agentregistry-dev/agentregistry/pkg/registry/v1alpha1store"
+)
+
+const (
+	// ReconcileActionApply converges a Deployment toward running desired state.
+	ReconcileActionApply = "apply"
+	// ReconcileActionRemove tears down runtime resources.
+	ReconcileActionRemove = "remove"
```
shashankram: Should we use `Delete/delete` instead of `remove` to be consistently with `arctl/API delete` terminlogy?

internal/registry/controller/deployment.go
```diff
@@ -0,0 +1,89 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/agentregistry-dev/agentregistry/pkg/api/v1alpha1"
+	"github.com/agentregistry-dev/agentregistry/pkg/registry/v1alpha1store"
+)
+
+const (
+	// ReconcileActionApply converges a Deployment toward running desired state.
+	ReconcileActionApply = "apply"
```
shashankram: Let's make this an Enum type

internal/registry/controller/deployment.go
```diff
@@ -0,0 +1,89 @@
...
+	ReconcileActionRemove = "remove"
+)
+
+// DeploymentRequestPayload is the payload persisted with Deployment
+// reconcile_work. It captures refs needed by a future executor without
+// embedding full object payloads.
+type DeploymentRequestPayload struct {
+	TargetRef    v1alpha1.ResourceRef `json:"targetRef"`
+	RuntimeRef   v1alpha1.ResourceRef `json:"runtimeRef"`
+	DesiredState string               `json:"desiredState,omitempty"`
```
shashankram: Should be an enum

internal/registry/controller/deriver.go
```diff
@@ -0,0 +1,118 @@
...
+		return nil
+	}
+	intents := d.workIntents()
+	if intents == nil {
+		return nil
+	}
+	return []krt.HandlerRegistration{
+		intents.RegisterBatch(func(events []krt.Event[DeploymentWorkIntent]) {
+			if _, err := d.DeriveIntentEvents(ctx, events); err != nil {
+				slog.Error("deployment controller derivation failed", "error", err)
```
shashankram: let's create a pkg scoped logger and use that instead of the global logger.
```
var logger = logging.New("registry-controller") 
```

internal/registry/controller/runner.go
```diff
@@ -0,0 +1,174 @@
...
+	internaldb "github.com/agentregistry-dev/agentregistry/internal/registry/database"
+	"github.com/agentregistry-dev/agentregistry/pkg/api/v1alpha1"
+	"github.com/agentregistry-dev/agentregistry/pkg/registry/v1alpha1store"
+	"github.com/agentregistry-dev/agentregistry/pkg/types"
+)
+
+const (
+	defaultProjectorResyncInterval = time.Minute
+	defaultExecutorInterval        = time.Second
+	defaultExecutorBatchLimit      = 10
```
shashankram: Why is this value chosen?. It seems like defaultExecutorInterval and defaultExecutorBatchLimit determine the throughput, but I am unsure how these values determine that and how can they be tuned.

internal/registry/controller/controller.go
```diff
@@ -0,0 +1,371 @@
...
+		}
+		opts.Cursor = cursor
+	}
+}
+
+func (c *DeploymentController) upsertDeploymentWork(ctx context.Context, deployment *v1alpha1.Deployment) error {
+	if c == nil || c.Work == nil {
+		return errors.New("deployment controller: reconcile work store is required")
+	}
+	work, err := DeriveDeploymentWork(deployment)
```
timflannagan: Does DeriveDeploymentWork need to be exported?

internal/registry/api/router/v0.go
```diff
@@ -175,48 +166,17 @@ func registerKindRoutes(
 	if registryValidator == nil {
 		registryValidator = registries.Dispatcher
 	}
-	// When a Deployment coordinator is supplied, install its Apply/Remove
```
timflannagan, line 178: 👍 makes sense making the executor the brain instead of embedding a coordinator in the REST server.

internal/registry/service/deployment/adapter_resolver.go
timflannagan, line 1: Rename this file?

internal/registry/service/deployment/adapter_resolver.go
```diff
@@ -2,242 +2,85 @@ package deployment
...
-//     annotations into Deployment.Metadata.Annotations.
-//  4. surface a structured error when no adapter is registered for a
-//     runtime's platform.
-//
-// Coordinator is NOT responsible for Upserting the Deployment row — that
-// happens upstream at the apply handler. Coordinator.Apply MUST be called
-// only after the row is persisted so status writes land on a real row.
-type Coordinator struct {
-	stores   map[string]*v1alpha1store.Store
+// AdapterResolver resolves Deployment runtime adapters for adjacent operations
```
timflannagan, line 13: A bit confused on the purpose for this interface after ripping out the coordinator. Just read through the code, this is now responsible for the Logs() and Discover() interfaces; apply/delete semantics were moved to the executor. Where I'm getting confused -- they both define resolveAdapter, resolveRuntime, refNamespace, etc. Should the reconciler own ref resolution + adapter lookup?

internal/registry/registry_app.go
```diff
@@ -88,18 +90,23 @@ func App(ctx context.Context, opts ...types.AppOptions) error {
...
-	// at admission so the coordinator's lookup can use exact-match.
+	// at admission so adapter lookup can use exact-match.
 	deploymentAdapters := map[string]types.DeploymentAdapter{
 		v1alpha1.TypeLocal:      local.NewLocalDeploymentAdapter(cfg.RuntimeDir, cfg.AgentGatewayPort),
 		v1alpha1.TypeKubernetes: kubernetes.NewKubernetesDeploymentAdapter(),
 	}
 	maps.Copy(deploymentAdapters, options.DeploymentAdapters)
 	pool := db.Pool()
 	stores := buildStores(pool, options.V1Alpha1StoreTables, options.V1Alpha1MutableStoreKinds, options.Auditor)
+	perKindHooks := crudPerKindHooks(options)
```
timflannagan: nit: define this closer to where it's used or inline directly in buildRouteOptions since that's the only call site for this variable.

pkg/api/v1alpha1/doc.go
timflannagan, line 1: These KindRegistry refactor changes are independent of the controller changes being made here, right?

timflannagan (commented):
I'd trying to understand whether we have the opportunity to further simplify these changes a bit after we ripped out KRT from the initial PR's diff. 

One opportunity that came to mind while looking at these changes again: simplifying the controller/executor components a bit into a single loop. You'd still have the control_plane_events table that acts as the indirect bridge/"contract" between the API server & deployment reconciler. That all still makes sense to me.

I think we could maintain an in-memory cache for maintaining reconciler work state, which could replace the DB related tables, right? Effectively, workqueue.RateLimitingInterface from the client-go/c-r world. Leader election can handle the multi-replica writes scenario.

The other thing that came to mind. Do we need a config field for backoffs at all?

internal/registry/controller/controller.go
ilackarms: Good catch. It does not need to be exported. Fixed in `33f181e6` by making it package-private as `deriveDeploymentWork` and updating the usages/tests.

internal/registry/service/deployment/adapter_resolver.go
ilackarms: Agreed. Renamed this to `adapter_resolver.go` and the matching test file to `adapter_resolver_test.go` in `33f181e6`.

internal/registry/service/deployment/adapter_resolver.go
ilackarms: The intended boundary is: apply/delete ref resolution and adapter lookup are owned by the controller executor because those paths lead to external side effects and must sit behind durable work claims. `AdapterResolver` is only for the synchronous adjacent read/inspect paths (`Logs` and `Discover`) and never calls `Apply`/`Remove`.

I renamed the stale coordinator file in `33f181e6` to make that less confusing. I did not route logs/discover through the executor because they are not durable reconcile side effects, but we can still extract the tiny shared runtime/adapter lookup helpers later if the duplication becomes noisy.

internal/registry/registry_app.go
ilackarms: Done in `33f181e6`; `crudPerKindHooks(options)` is now inlined at the `buildRouteOptions` call site.

pkg/api/v1alpha1/doc.go
ilackarms: Yes, mostly support refactor rather than controller behavior by itself. It keeps v1alpha1 scheme/store/routes/plurals/table metadata in one registry so the controller and generic resource surfaces do not keep separate hard-coded kind lists.

This could be split out mechanically, but keeping it here avoids a second dependency PR while this controller foundation needs consistent kind metadata and event coverage across the built-in kinds.

timflannagan (approved):
Thanks for sticking with this after multiple rounds of major refactors compared to the original PR. I do think the removal of KRT and some of the reconcile_* DB tables help simplify these changes quite a bit and make it easier to reason about the server architecture works going forward.

I still think there's some opportunity to define the contracts a bit more between each system, but having a dedicated Deployment controller is moving the needle quite a bit here and I'd like to see where this work lands in the near term before forming any strong opinion. And obviously this work unblocks important functional work, e.g. drift detection. Won't bother to flag inline nit review feedback as a result of that.

One thing I didn't run down but I'm still confused about: what's the plan with the RuntimeAdapter now that the deployment adapters have moved async? We can talk offline as well, just want to understand the order of operations after we land these foundational changes.

An alternative approach I still think is worth exploring. Leveraging controller-runtime's primitives around NewUnmanaged + source.Channel, which allows us to inherit quite a bit of functionality we'll have to define natively in the server in the short/medium term.

PR #523, opened 2026-05-29
helm: add database.postgres.secretRef to source the URL from a Secret

xytian315:
# Description
Adds `database.postgres.secretRef.{name, key}` to the agentregistry chart so users can source the Postgres connection URL from an existing Kubernetes Secret instead of inlining it in values.

# Change Type

/kind feature
# Changelog

```release-note
feat(helm): support sourcing the Postgres connection URL from an existing Kubernetes Secret via `database.postgres.secretRef.{name,key}`
```

# Additional Notes

**Two accepted connection-string formats.** `AGENT_REGISTRY_DATABASE_URL` accepts either libpq URL form
  (`postgres://user:pass@host:port/db?sslmode=require`) or keyword/value form (`host=... user=... password='...' dbname=... sslmode=require`). The pgx driver auto-detects based on the leading `postgres://` prefix. Keyword/value form is recommended when credentials come from a rotating secret store (AWS Secrets Manager, Vault) — it avoids URL-encoding special characters in the password on every rotation.


timflannagan (approved):

timflannagan:
Just removing from the MQ since Nik mentioned he can take a look too

nikolasmatt (approved):
LGTM!
noting that this is intentionally different than the `urlFile` approach [kagent](https://github.com/kagent-dev/kagent/blob/main/helm/kagent/values.yaml#L72-L74) taken. IMO this is a simpler and more standard interface+implementation.
cc: @iplay88keys @EItanya

PR #530, opened 2026-06-03
[BREAKING] feat(api): MCPPackage v1alpha1 redesign (Origin/Launch/Transport)

xytian315:
# Description

  **What changed:**

  - **MCPPackage v1alpha1 redesign (breaking).** Replaces the flat `MCPPackage` schema with a polymorphic shape: `Origin.{Type, Identifier, NPM|PyPI|OCI}` + `Launch.{Command, Args, Env}` (nullable) + `Transport`. Per-origin validators dispatch on `MCPPackageOrigin`; the top-level validator enforces exactly-one-of polymorphism. OCI origin requires explicit `tag` or `digest`. `nuget` and `mcpb` registry types are dropped entirely. Existing `mcp.yaml` files need to migrate — see updated `examples/mcp.yaml` and `examples/full-stack.yaml`.

  - **`arctl init mcp --transport {stdio,http}` (default http).** Frameworks declare per-transport launch defaults in `framework.yaml` (`launch: {stdio: {...}, http: {...}}`); `arctl init mcp` writes the matching block into the scaffolded `mcp.yaml`'s `spec.source.package.launch`. HTTP args use Go text/template substitution for `{{.Port}}`. Fixes a real bug where HTTP source-built MCPs would crash-loop because the container `ENTRYPOINT` defaulted to stdio.

  - **Dropped the hard-coded `MCP_TRANSPORT_MODE` env-var convention.** The scaffolded fastmcp template no longer reads `os.getenv("MCP_TRANSPORT_MODE", ...)`; its argparse `--transport` default flipped from `stdio` to `http`. `MCPPackage.Transport.Type` is the sole  source of truth for transport mode.

  - **ADK tool-name prefixing in scaffolded agents.** `arctl init agent`'s scaffolded `mcp_tools.py` constructs each `MCPToolset(...,  tool_name_prefix=_tool_name_prefix(server_name))`. Resolves LLM API 400s ("tools: Tool names must be unique") when multiple MCPs expose tools with the same name.

  - **mcp-go gets `localLaunch` / `localInstall`.** Framework descriptors declare a `localLaunch` block (the command `arctl run --transport stdio` executes in-place) and `localInstall` (e.g. `go mod tidy`). Lets you iterate on Go-based stdio MCPs locally without rebuilding the container image.

# Change Type

/kind breaking_change
/kind cleanup
/kind feature
# Changelog

```release-note
BREAKING: MCPPackage v1alpha1 redesigned (`Origin` + `Launch` + `Transport`); New: `arctl init mcp --transport {http,stdio}` and per-server `tool_name_prefix` in scaffolded `mcp_tools.py`
```

# Additional Notes


internal/cli/declarative/init.go
```diff
@@ -57,6 +58,19 @@ func resolveInitProjectPath(cmd *cobra.Command, projectName string) (string, err
...
+// resolveInitTransport finalizes the MCP transport for `arctl init mcp`:
+// uses the explicit --transport flag if set, otherwise defaults to "http".
+// Rejects stdio + --port.
+func resolveInitTransport(cmd *cobra.Command, current string) (string, error) {
+	if current == "" {
+		current = "http"
+	}
+	if current == "stdio" && cmd.Flags().Changed("port") {
+		return "", fmt.Errorf("--port is meaningless with --transport stdio")
+	}
```
timflannagan: nit: Why is this logic required? We check this in the inline RunE before running this function

internal/cli/declarative/init.go
timflannagan: nit: I think we can just rip out the appendLocalMCPsToDotEnv mention since that code no longer exists?

internal/cli/declarative/init.go
```diff
@@ -449,9 +463,19 @@ func localMCPEntries(paths []string) ([]mcpEnvEntry, error) {
...
+			return nil, fmt.Errorf("sibling MCP at %s declares transport.type=stdio; "+
+				"--local-mcp wires siblings via an HTTP URL, so only http-transport MCPs are supported", abs)
+		}
 		port := cfg.Port
 		if port == 0 {
-			port = 3000
+			return nil, fmt.Errorf("sibling MCP at %s has no port in arctl.yaml; "+
+				"re-scaffold with `arctl init mcp --transport http --port <PORT>`", abs)
 		}
 		siblingName, err := readMCPName(abs)
```
timflannagan: nit: Looks like we serialize the MCPServer twice now? readMCPName calls readMCPYAML under the hood. I think we can remove readMCPName and just use the doc.Metadata.Name now?

internal/cli/declarative/init.go
```diff
@@ -560,11 +584,7 @@ func loadFrameworkRegistry(projectRoot string) (*frameworks.Registry, error) {
 }
 
 func isatty() bool {
-	fi, err := os.Stdin.Stat()
-	if err != nil {
-		return false
-	}
-	return (fi.Mode() & os.ModeCharDevice) != 0
+	return term.IsTerminal(int(os.Stdin.Fd()))
```
timflannagan, line 553: Why is this change needed here?

internal/cli/frameworks/builtin/fastmcp-python/templates/Dockerfile.tmpl
```diff
@@ -60,14 +60,13 @@ EXPOSE 8080
 
 # Health check
 HEALTHCHECK --interval=30s --timeout=10s --start-period=5s --retries=3 \
-    CMD python -c "import src.main; print('healthy')"
+    CMD python3 -c "import src.main; print('healthy')"
```
timflannagan, line 63: Just curious -- why is the python3 change necessary here?

internal/cli/frameworks/builtin/adk-python/templates/{{.Name}}/mcp_tools.py.tmpl
timflannagan, line 1: Why are these changes necessary here? Just a gap we identified while digging into the redesign? Regardless, it seems like a bug with the existing code scaffolding, just wanted to understand whether I was missing any context of this change.

internal/cli/declarative/inspector/inspector.go
timflannagan, line 1: I think we'll need to refactor this package at some point. Extending in the current shape is fine for now though.

pkg/api/v1alpha1/registries/oci.go
```diff
@@ -59,25 +55,18 @@ var allowedOCIRegistries = map[string]bool{
 //   - Docker Hub (docker.io)
 //   - GitHub Container Registry (ghcr.io)
 //   - Google Artifact Registry (*.pkg.dev)
-func ValidateOCI(ctx context.Context, pkg v1alpha1.RegistryPackage, serverName string) error {
-	if pkg.Identifier == "" {
-		return ErrMissingIdentifierForOCI
+func ValidateOCI(ctx context.Context, origin v1alpha1.MCPPackageOrigin, serverName string) error {
```
timflannagan, line 58: Unrelated to these changes. Don't think we need to export ValidateOCI anymore?

pkg/api/v1alpha1/registries/oci.go
```diff
@@ -59,25 +55,18 @@ var allowedOCIRegistries = map[string]bool{
...
-		return fmt.Errorf("OCI packages must not have 'registryBaseUrl' field - use canonical reference in 'identifier' instead (e.g., 'docker.io/owner/image:1.0.0')")
-	}
-	if pkg.Version != "" {
-		return fmt.Errorf("OCI packages must not have 'version' field - include version in 'identifier' instead (e.g., 'docker.io/owner/image:1.0.0')")
+	if origin.Identifier == "" {
+		return ErrMissingIdentifierForOCI
 	}
-	if pkg.FileSHA256 != "" {
-		return fmt.Errorf("OCI packages must not have 'fileSha256' field")
+	if !ociIdentifierHasTagOrDigest(origin.Identifier) {
```
timflannagan, line 65: Need to catch up a bit with this code vs. the mcpserver_validate.go code. Some of these checks feel duplicated across these packages?

internal/cli/frameworks/builtin/fastmcp-python/templates/Dockerfile.tmpl
timflannagan: We expose a --port flag but hardcode in the dockerfile? Didn't dig too deep, not sure if this is an actual issue

pkg/api/v1alpha1/mcpserver.go
```diff
@@ -84,81 +90,196 @@ type MCPTransport struct {
...
-}
-
-// MCPArgument.Type values. Kept as string literals to match the YAML wire
-// format; platform translators compare against these.
+	Origin    MCPPackageOrigin  `json:"origin" yaml:"origin"`
+	Launch    *MCPPackageLaunch `json:"launch,omitempty" yaml:"launch,omitempty"`
+	Transport MCPTransport      `json:"transport" yaml:"transport"`
+}
+
+// MCPPackageOrigin identifies the package and where to fetch it. The Type
```
timflannagan, line 96: Talked offline a bit. Just bubbling up my 2c on the new API shape. Agreed on making the MCPPackage domain oriented. My ideal world is that we refactor MCPPackageOrigin again at some point where it's grouped by packaging domain, e.g. oci, registry (npm, pypi, etc.) are grouped together instead of being top-level types, expand to artifact as a supported type for the mcpb use case we're ripping out here.

Additionally, the new, top-level "identifier" field is awkward to me in the current shape and may cause confusion to users. Already discussed this concern offline, we'll move forward with these changes, etc.

pkg/api/v1alpha1/mcpserver.go
```diff
@@ -84,81 +90,196 @@ type MCPTransport struct {
...
+//
+//	args: [src/main.py, "--debug"]
+//
+// Each string element of the shorthand becomes a positional MCPArgument
+// with Value set to the string. Mixing string scalars with mapping
+// elements in the same list is rejected with a clear error.
+//
+// On marshalling, MCPPackageLaunch always emits the structured form;
+// the shorthand is an input-only convenience.
+func (l *MCPPackageLaunch) UnmarshalYAML(value *yaml.Node) error {
```
timflannagan: Confused on why we need a UnmarshalYAML method here. I don't see this pattern used through any of our API type definitions

pkg/api/v1alpha1/mcpserver.go
```diff
@@ -84,81 +90,196 @@ type MCPTransport struct {
...
+		l.Args = args
+		return nil
+	default:
+		return fmt.Errorf("launch.args[0]: expected a string scalar or a mapping, got %s", nodeKindName(r.Args.Content[0].Kind))
+	}
+}
+
+// nodeKindName returns a short label for a yaml.Node kind, used in error
+// messages so users see "mapping" / "sequence" rather than raw integers.
+func nodeKindName(k yaml.Kind) string {
```
timflannagan: Same question as above. Why do we need this logic? Does this belong in mcpserver_validate.go instead?

timflannagan (approved):

internal/cli/frameworks/builtin/fastmcp-python/templates/Dockerfile.tmpl
xytian315: this is to get  `arctl run <mcp>`  working locally for macOS system without frictions.

internal/cli/frameworks/builtin/adk-python/templates/{{.Name}}/mcp_tools.py.tmpl
xytian315: yes, I noticed this problem when updating downstream walkthrough. When multiple MCPs in the same agent expose tools with the same name, ADK errors out with `"tools: Tool names must be unique"`  This `tool_name_prefix` change namespaces each MCPToolset by server so collisions don't happen. 
You will notice this in the output changed in the walkthrough logs.

internal/cli/declarative/init.go
xytian315: removed

internal/cli/declarative/init.go
xytian315: fixed

internal/cli/declarative/init.go
xytian315: fixed.

pkg/api/v1alpha1/mcpserver.go
xytian315: dropped .

timflannagan (approved):

PR #534, opened 2026-06-05
ci: drop redundant caches and add timeout in build workflow

### LINKED ISSUE #469: Refactor the build.yml GHA workflow
Placeholder: https://github.com/agentregistry-dev/agentregistry/blob/main/.github/workflows/build.yml.

devc007:
# Description

Refactor build.yaml: setup-go/setup-node already handle module and
npm caching, so the two manual actions/cache@v3 steps were redundant.
Removed them and added a job timeout. Fixes #469.

# Change Type

/kind cleanup

# Changelog

```release-note
NONE
```

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -12,6 +12,7 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
+    timeout-minutes: 30
 
     steps:
     - uses: actions/checkout@v6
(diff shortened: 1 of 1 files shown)


devc007:
PTAL @timflannagan

timflannagan:
What I had in mind (and the issue description was too vague, mb) is that the build.yaml workflow provides very little value right now and should be deflated into existing workflows

PR #535, opened 2026-06-05
Restore unmanaged deployment discovery in lists

ilackarms:
# Description

Restores unmanaged Deployment discovery in list responses. Deployment adapters already expose out-of-band workloads through `Discover`, but the generic list path no longer called that contract, so the UI/API could not surface resources that existed on a registered runtime but were not registry-managed.

What changed:
- Added a generic list augmentation hook for read-only synthetic rows.
- Wired Deployment list augmentation to runtime adapter discovery.
- Added `origin=managed|discovered` list filtering.
- Added integration coverage proving managed rows dedupe and unmanaged rows appear as `origin=discovered`.
- Hardened managed-row dedupe to also treat provider `remoteName` / `remoteId` annotations as managed identities, which prevents sanitized runtime names from reappearing as unmanaged discoveries.
- Clarified Deployment CLI delete comments/tests: explicit delete is keyed by Deployment `metadata.name`, not shared `spec.targetRef.name`.

# Change Type

/kind fix

# Changelog

```release-note
Deployments discovered out-of-band by runtime adapters are now included in deployment list responses and can be filtered with origin=discovered.
```

# Additional Notes

Companion enterprise PR: https://github.com/solo-io/agentregistry-enterprise/pull/804

Validation:
- `go test ./internal/cli/declarative`
- `go test ./internal/registry/controller`
- `go test -tags=integration ./internal/registry/api/router -run TestDeploymentListMergesDiscoveredRows -count=1`
- `go test -tags=integration ./internal/registry/controller -run 'TestDeploymentController_(BlocksMissingTargetWithoutAdapterCall|ReappliesWhenMissingTargetAppears|SkipsUnchangedApplyAfterRepairReconcile)' -count=1`


shashankram:
Does List for Deployments show unmanaged/discovered deployments by default now? The CLI should not show those by default, there should be a filter to list them if required. cc @Charlesthebird for UI

pkg/cli/runtime/runtime.go
```diff
@@ -72,10 +74,16 @@ func (r *runtime) RegistryClient(ctx context.Context) (*client.Client, error) {
...
+			token, err := r.cfg.Auth.Token(ctx)
+			if errors.Is(err, types.ErrCLINoStoredToken) {
+				token = ""
+				err = nil
+			}
+			if err != nil {
+				r.clientErr = err
 				return
 			}
+			target.Token = token
```
timflannagan, line 86: Why was this change necessary? Seems unrelated to the discovery fixes?

internal/registry/api/router/deployment_discovery.go
```diff
@@ -0,0 +1,233 @@
...
+	return func(ctx context.Context, in resource.ListAugmentInput) ([]*v1alpha1.RawObject, error) {
+		if discoverer == nil || in.Origin == "managed" || in.Cursor != "" || in.Labels != "" {
+			return nil, nil
+		}
+		runtimeStore := stores[v1alpha1.KindRuntime]
+		if runtimeStore == nil {
+			return nil, nil
+		}
+
+		runtimes, _, err := runtimeStore.List(ctx, v1alpha1store.ListOpts{
```
timflannagan: Fine for now since registry managed runtimes will be a low N in the short term, but something we may need to optimize in the long term.

pkg/registry/resource/handler.go
```diff
@@ -717,6 +792,30 @@ func runList[T v1alpha1.Object](
...
+		}
+		if p.Origin == "discovered" {
+			rows = extra
+		} else {
+			rows = append(rows, extra...)
+		}
+	}
+	if p.Origin == "discovered" {
+		nextCursor = ""
+	}
```
timflannagan: What's the context for this change? No need for pagination for discovered resources?

internal/registry/api/router/deployment_discovery.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+	for _, key := range []string{"remoteName", "remoteId"} {
+		if value := strings.TrimSpace(result.RuntimeMetadata[key]); value != "" {
+			return value
+		}
+	}
+	return ""
+}
+
+func discoveredDeploymentName(runtimeName, targetKind, targetName, tag, namespace string) string {
```
timflannagan: This hashing logic makes sense in terms of producing unique & deterministic names, but at the cost that it's produces opaque names in the UI, right? I guess this is fine for now to help unblock the discovery fixes.

pkg/registry/resource/handler.go
```diff
@@ -269,6 +280,35 @@ type listInput struct {
 	IncludeTerminating bool `query:"includeTerminating" doc:"Include rows with a deletionTimestamp."`
 }
 
+type listWithOriginInput struct {
```
timflannagan, line 283: This is the same type as the listInput with the Origin field added? Wonder if we can just embed listInput here instead instead of maintaining duplicate fields across both structs. Not sure if Huma supports embedding though

timflannagan (commented):
LGTM, just one quick question about the auth token changes here.

pkg/registry/resource/handler.go
ilackarms: Context is that discovered rows are synthetic live adapter results, not DB rows, so the DB cursor cannot accurately page them. For `origin=discovered` we return the current discovery snapshot and clear `nextCursor`; if a caller supplies `cursor` or `labels`, the augmenter opts out to avoid mixing a DB cursor with provider-derived rows. Managed rows keep normal pagination. A real paginated discovered listing would need a materialized discovery store or adapter pagination token.

internal/registry/api/router/deployment_discovery.go
ilackarms: Yes, that is the tradeoff. The registry `metadata.name` is an opaque deterministic ID so synthetic rows are safe and unique across namespace/runtime/target kind/target/tag, and stable across list calls. The human-readable pieces still live in `spec.targetRef.name`, `spec.runtimeRef.name`, and `status.details.runtimeMetadata` for UI display. If the UI needs a friendlier `metadata.name`, we can evolve this to a readable prefix plus hash suffix later.

pkg/cli/runtime/runtime.go
ilackarms: Good question. This came up while exercising the CLI against authless/dev registry paths: `RegistryClient` already documents best-effort client creation, but `ErrCLINoStoredToken` made construction fail locally before the request reached the server. Treating only that sentinel as an empty token lets unauthenticated calls reach the registry and preserves real auth-provider failures. It is adjacent rather than core discovery logic, so I can split it out if you prefer, but it fixes the CLI path used to verify these list/discovery flows.

internal/registry/api/router/deployment_discovery.go
ilackarms: Agreed. This is intentionally simple for now: discovery scans visible Runtime rows with a cap of 1000 and calls adapter discovery per supported runtime. In the near term N should be small, and unsupported runtime types are skipped. Longer term, if runtime cardinality grows, we should move this to materialized discovery/cache or adapter-scoped runtime filtering/pagination instead of list-time fanout.

pkg/registry/resource/handler.go
ilackarms: Good call. I checked Huma v2 and it does recurse through embedded structs for input params, so this should be safe to collapse to an embedded `listInput` plus the `Origin` field. I will make that cleanup and keep `TestResourceRegister_OriginFilterIsOptIn` as the guard that `origin` remains opt-in for deployments only.

pkg/registry/resource/handler.go
ilackarms: I think, looking back at it, I would rather just persist all of our deployments to the DB rather than using the synthetic in-memory adapter that we use today. My plan was to keep only a single source of truth in the DB, but it's part of what blocks us from being able to do pagination here, because discovered v1alpha1 deployments don't get written into the DB today. I think I would prefer to have our adapter use the db table and then I can simplify this logic a bit.

timflannagan (approved):

PR #549, opened 2026-06-13
Support GitLab skill source URLs

### LINKED ISSUE #548: Support GitLab repositories for skill publish and skill pull
## Summary

`arctl` can list Git-backed skills whose registry records point to GitLab URLs, but `arctl skill pull` fails because the CLI only supports GitHub URL parsing.

## Environment

- agentregistry / arctl: v0.3.3
- Registry running at: http://100.77.92.37:12121
- Skill source: private GitLab instance

## Reproduction

Create or register a skill record with a GitLab source URL:

```json
{
  "name": "ticket-debug",
  "description": "Ticket debugging workflow from GitLab",
  "version": "1.0.0",
  "repository": {
    "source": "git",
    "url": "https://gitlabe2.ext.net.org.com/mansoora/pam/-/blob/main/skills/ticket_debug.md"
  }
}

Then run:

arctl skill pull ticket-debug .cursor/skills/ticket-debug \
  --registry-url http://100.77.92.37:12121 \
  --version 1.0.0
Actual Result
Error: parse GitHub URL: unsupported host "gitlabe2.ext.net.org.com", only github.com is supported
Also, arctl skill publish --git <gitlab-url> rejects GitLab URLs with the same GitHub-only host validation.

Expected Result
arctl skill publish --git and arctl skill pull should support GitLab repository URLs, including self-hosted GitLab instances.

Useful supported forms could include:

https://gitlab.example.com/group/project/-/blob/main/path/to/SKILL.md
https://gitlab.example.com/group/project/-/tree/main/path/to/skill
https://gitlab.example.com/group/project.git
For private GitLab repos, arctl should either:

use existing git credentials,
support PAT/token configuration,
or clearly document the supported auth flow.
Why This Matters
Many enterprise teams host skills in private GitLab, not GitHub. The registry UI/API can store GitLab-backed skill records, but the CLI cannot pull them into Cursor/local workspaces, which breaks the advertised discover/pull workflow.

To create with `gh` after installing/authenticating:
```bash
gh issue create \
  --repo agentregistry-dev/agentregistry \
  --title "Support GitLab repositories for skill publish and skill pull" \
  --body-file issue.md

### PR DESCRIPTION
**Pull request by mansoor17syed**
# Description

Adds GitLab and self-hosted GitLab support for Git-backed skill source URLs.

Previously, `arctl` parsed skill source repositories through GitHub-specific logic, so GitLab-backed skills registered in the registry could not be pulled locally. This change introduces generic Git URL parsing while preserving existing GitHub behavior.

What changed:
- Added `ParseGitURL` for generic Git web URLs.
- Kept `ParseGitHubURL` as a compatibility wrapper.
- Added support for GitLab `/-/tree/...` and `/-/blob/...` URLs.
- Added support for self-hosted GitLab instances.
- Allows file subpaths, such as a direct `SKILL.md` blob URL, to be copied into the target directory.
- Updated Kubernetes runtime Git skill parsing to use the generic parser.

Fixes #548

# Change Type

/kind feature
/kind fix

# Changelog
```release-note
Add support for GitLab and self-hosted GitLab URLs when pulling Git-backed skill sources.
```


internal/cli/common/gitutil/gitutil.go
```diff
@@ -21,33 +21,49 @@ import (
 // encoded as %2F in the URL. The raw (escaped) path is used for splitting so
 // the encoded branch segment is preserved, then unescaped for the return value.
 func ParseGitHubURL(rawURL string) (cloneURL, branch, subPath string, err error) {
```
lcopi: this wrapper isn't providing any value so I think we can go ahead and get rid of it

internal/cli/common/gitutil/gitutil.go
```diff
@@ -21,33 +21,49 @@ import (
...
-	// Path is like /owner/repo or /owner/repo/tree/branch/sub/path
+	// Path is like /owner/repo, /owner/repo/tree/branch/sub/path, or
+	// /group/project/-/tree/branch/sub/path for GitLab.
 	parts := strings.Split(strings.Trim(rawPath, "/"), "/")
 	if len(parts) < 2 {
-		return "", "", "", fmt.Errorf("invalid GitHub URL: expected at least owner/repo in path")
+		return "", "", "", fmt.Errorf("invalid Git URL: expected at least namespace/repo in path")
+	}
+
+	if gitLabMarker := indexPart(parts, "-"); gitLabMarker >= 2 {
```
lcopi: looking for an explicit gitlab in the URL to use as a fallthrough to that parser in addition to the URL index parsing would be a good optimization.

internal/cli/common/gitutil/gitutil.go
```diff
@@ -162,6 +205,18 @@ func CopyRepoContents(repoDir, subPath, targetDir string) error {
 		if err != nil {
 			return err
 		}
+		if info, err := os.Lstat(resolved); err != nil {
+			return fmt.Errorf("stat subpath %q: %w", subPath, err)
+		} else if !info.IsDir() {
+			if err := os.MkdirAll(targetDir, 0o755); err != nil {
```
lcopi: this should probably move to after the symlink check so it doesn't orphan directories on disk

internal/cli/common/gitutil/gitutil.go
```diff
@@ -21,33 +21,49 @@ import (
...
+// GitLab instances that use /-/tree/ and /-/blob/ routes.
+func ParseGitURL(rawURL string) (cloneURL, branch, subPath string, err error) {
 	u, err := url.Parse(rawURL)
 	if err != nil {
 		return "", "", "", fmt.Errorf("invalid URL: %w", err)
 	}
 
-	if u.Host != "github.com" {
-		return "", "", "", fmt.Errorf("unsupported host %q, only github.com is supported", u.Host)
+	if u.Scheme == "" || u.Host == "" {
```
lcopi: scheme should probably be a bit more restrictive here

internal/cli/common/gitutil/gitutil.go
lcopi: You should be able to get rid of the `indexPart` helper and use `slices.Index(parts, "-")` instead

lcopi (changes requested):

mansoor17syed:
Thanks for the review. I pushed a follow-up commit addressing the comments:

- removed the ParseGitHubURL wrapper
- replaced indexPart with slices.Index
- restricted supported URL schemes to http/https
- added explicit GitLab host fallback, including nested group repo roots
- moved target directory creation until after symlink rejection
- added regression coverage for unsupported schemes and symlink file subpaths

Tested with:
go test ./internal/cli/common/gitutil ./internal/registry/runtimes/kubernetes

lcopi (approved):

mansoor17syed:
@lcopi 
Could you please review this and assist with the merge?

PR #550, opened 2026-06-15
feat(api): read-only MCP Registry v0.1 compatibility endpoints

ilackarms:
# Description

**Motivation.** AgentRegistry began as a fork of the [official MCP Registry](https://github.com/modelcontextprotocol/registry) but reorganized its public API around Kubernetes-style `v1alpha1` resources (`MCPServer` at `/v0/mcpservers`). That moved it off the upstream `server.json` contract, so registry-aware clients — notably VS Code's MCP gallery — can no longer consume AgentRegistry as an MCP registry.

**What changed.** This adds a **read-only** compatibility shim that re-projects existing `MCPServer` resources into the **frozen MCP Registry v0.1** `server.json` shape and serves the spec's standard discovery paths. The native `v1alpha1` API is unchanged and remains the source of truth; this is purely additive.

Endpoints (served at root, or under an optional configured prefix):
- `GET /v0.1/servers` — list, cursor-paginated; supports `cursor`, `limit` (≤100), `search`, `updated_since`, `version`, `include_deleted`
- `GET /v0.1/servers/{serverName}/versions` — list versions of one server
- `GET /v0.1/servers/{serverName}/versions/{version}` — get one version (`{version}` accepts `latest`)

Responses match the upstream envelope verbatim: `{ servers: [{ server, _meta }], metadata: { nextCursor, count } }`, camelCase fields, and the `io.modelcontextprotocol.registry/official` `_meta` block, pinned to the `2025-12-11` schema.

**Design notes:**
- **Flattened, anonymous catalogue.** The catalogue spans all namespaces; each server's `name` is `"<namespace>/<resourceName>"` (one slash, unique, reversible for the get-by-name routes). Reads are anonymous, matching the upstream public-read model.
- **Layering.** New `pkg/mcpregistry` owns the v0.1 types + the `v1alpha1 → server.json` translation; a transport-only handler in `internal/registry/api/handlers/mcpregistry` reads the existing `MCPServer` store and maps errors. Wired in `RegisterRoutes` (a first-party OSS read feature), gated by config.
- **Standard paths, no collision.** `/v0.1/servers` doesn't collide with the native `/v0/*` surface, so clients can use the registry root as their base URL; an optional `…_PATH_PREFIX` covers prefixed mounts. VS Code points its enterprise `McpGalleryServiceUrl` policy at that base URL.
- **Config.** `AGENT_REGISTRY_MCP_REGISTRY_COMPAT_ENABLED` (default `false`, opt-in) and `AGENT_REGISTRY_MCP_REGISTRY_COMPAT_PATH_PREFIX` (default empty). See `docs/mcp-registry-compatibility.md`.

> **Security posture — default OFF (opt-in), RBAC-aware.** The handler reuses the same per-kind `ListFilter` + `Authorize` hooks (`crud.PerKindHooks`) as the native MCPServer read path: the list applies `ListFilter`, single-server reads apply `Authorize` (forbidden → 404). OSS leaves them nil (flat, unfiltered catalogue — matches already-public OSS reads); a downstream build that gates MCPServer reads gets identical scoping with no extra wiring. Because the OSS default is unauthenticated + cross-namespace, it ships **disabled by default** (`AGENT_REGISTRY_MCP_REGISTRY_COMPAT_ENABLED`). Documented in the authz matrix. *(Enterprise companion: solo-io/agentregistry-enterprise#926 — draft, RBAC is automatic, pending this PR's merge for the dep bump.)*

Tests: table-driven translation tests (npm/pypi/oci, stdio/http transport, remote-only, args/env, name round-trip, version/isLatest derivation) and `httptest` handler tests (list envelope, pagination passthrough, filter mapping, `%2F`-encoded names, 404s). `openapi.yaml` and the generated TS client are regenerated.

# Change Type

/kind feature

# Changelog

```release-note
Add a read-only MCP Registry v0.1 compatibility API (GET /v0.1/servers ...) that re-exposes MCP servers in the official server.json format so registry-aware clients (e.g. VS Code) can discover them. Off by default; enable with AGENT_REGISTRY_MCP_REGISTRY_COMPAT_ENABLED=true.
```

# Additional Notes

- `search` matches the AgentRegistry resource-name column (substring), not the namespace-qualified catalogue name; `http` package transports are surfaced as `streamable-http` with a synthesized `http://localhost:<port><path>` URL. Both are documented as best-effort mappings.
- Verified locally: `go build ./...`, `go vet`, package tests, `golangci-lint` (0 issues), `gofmt`, `go mod tidy` (no drift), and OpenAPI + TS client regeneration.


ilackarms:
## Verified live against VS Code (1.124.0)

Beyond the unit/handler tests, I ran the shim end-to-end against a real VS Code MCP gallery, not a stub.

**Discovery contract.** VS Code's gallery calls the v0.1 API exactly as implemented — a version probe `GET /v0.1/servers?limit=1`, then the list `GET /v0.1/servers?limit=50&version=latest` (`Origin: vscode-file://…`). Its version candidates are `["v0.1","v0"]` and it tries **v0.1 first**, so **v0.1-only is sufficient** — no legacy `/v0` alias is needed.

**CORS.** The webview is a cross-origin caller; the server's existing permissive CORS handler (`Access-Control-Allow-Origin: *`, methods `GET, OPTIONS`, 204 preflight) satisfies the gallery's documented requirement. No change needed here.

**Registry validation (the real-server test).** Seeded the official `@modelcontextprotocol/server-everything` (npm / stdio). On apply it passed the npm ownership check — the validator fetches the package from the npm registry and requires its published `mcpName` to equal the declared `serverName` (`io.github.modelcontextprotocol/server-everything`). It was then discovered + installed from the VS Code gallery and a tool call succeeded from Copilot agent mode (`echo` → `Echo: hello from agentregistry`).

**Transport (deployment note, not a code concern).** VS Code's renderer CSP is `connect-src https:`, so the gallery URL must be served over **HTTPS** (`http://` is refused client-side). AgentRegistry runs behind TLS in production, so this is purely a deployment detail; VS Code is pointed at the registry via the `McpGalleryServiceUrl` device policy.

Net: a real VS Code can discover, install, and invoke MCP servers through the compatibility shim.

internal/registry/api/router/router.go
```diff
@@ -222,6 +222,7 @@ func NewHumaAPI(
 		mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
 			// Check if this is an API route - if so, return 404
 			if strings.HasPrefix(r.URL.Path, "/v0/") ||
+				strings.HasPrefix(r.URL.Path, "/v0.1/") ||
```
lcopi, line 225: the doc says these routes are unauthenticated, do they need to be added to the authn skip list?

internal/registry/api/handlers/mcpregistry/handlers.go
```diff
@@ -0,0 +1,277 @@
...
+			opts.LatestOnly = true
+		default:
+			opts.Tag = in.Version
+		}
+
+		preds := make([]string, 0, 2)
+		args := make([]any, 0, 2)
+		if in.Search != "" {
+			args = append(args, "%"+in.Search+"%")
+			preds = append(preds, fmt.Sprintf("name ILIKE $%d", len(args)))
```
lcopi, line 158: will clients pass in actual name here, or should this attempt to sanitize namespace off?

internal/registry/api/handlers/mcpregistry/handlers.go
```diff
@@ -0,0 +1,277 @@
...
+	Get(ctx context.Context, namespace, name, tag string) (*v1alpha1.RawObject, error)
+}
+
+var _ ServerStore = (*v1alpha1store.Store)(nil)
+
+// Register mounts the v0.1 compatibility routes on api. pathPrefix is prepended
+// to the standard `/v0.1` base (empty serves the spec paths at root); store is
+// the MCPServer store the endpoints read from.
+func Register(api huma.API, pathPrefix string, store ServerStore) {
+	base := pathPrefix + "/v0.1"
```
lcopi: does the environment variable input require a leading `/`, if someone sets `AGENT_REGISTRY_MCP_REGISTRY_COMPAT_PATH_PREFIX=registry` will that work or will it break the route registration?

pkg/mcpregistry/translate.go
```diff
@@ -0,0 +1,304 @@
...
+
+// packageTransportOf maps the deployable package transport. AgentRegistry's
+// "http" becomes the upstream "streamable-http" with a synthesized localhost
+// URL built from the listen port/path; "stdio" (and the empty default) map to
+// the stdio transport.
+func packageTransportOf(t v1alpha1.MCPTransport) ServerTransport {
+	if t.Type == "http" {
+		return ServerTransport{
+			Type: "streamable-http",
+			URL:  fmt.Sprintf("http://localhost:%d%s", t.Port, t.Path),
```
lcopi, line 216: is t.port guaranteed to be set? If its empty will the zero value cause issues?

internal/registry/api/router/router.go
ilackarms: Good Q — kept them out of the skip list on purpose. If we skip authn, no session reaches the RBAC `ListFilter`/`Authorize`, so in enterprise even a valid token would get an empty catalogue. OSS just doesn't configure an authn provider, so it's anonymous there. Clarified the wording in the router comment + authz matrix.

internal/registry/api/handlers/mcpregistry/handlers.go
ilackarms: Clients pass back the full `<namespace>/<name>` we emit, and `parseServerNameParam` splits it — the store query then uses `Namespace: ns` + `name = $1` (the bare name), so the namespace is parsed off and used for scoping.

internal/registry/api/handlers/mcpregistry/handlers.go
ilackarms: Good catch — `registry` would've broken ServeMux registration. Normalized now (prepend `/`, strip trailing `/`), so `registry` / `/registry` / `registry/` all land on `/registry/v0.1`. Added a test. (47f6a5de)

pkg/mcpregistry/translate.go
ilackarms: Port is required + non-zero for http transport (validated at apply — `mcpserver_validate.go:89`), so it's always set on stored servers; `localhost:0` can't happen. Added a comment calling that out. (47f6a5de)

lcopi (approved):

timflannagan (approved):

PR #554, opened 2026-06-22
feat(plugins): Plugin kind + harness agents (source-pointer storage)

ilackarms:
# Description

Adds the experimental **Plugin** kind and harness-based agents to AgentRegistry.

This PR implements the accepted source-pointer storage model without adding registry-hosted plugin bytes: a Plugin spec carries user intent for an external source, the controller resolves public git sources to a concrete commit out of band, and Plugin status records the resolved source pin plus parsed manifest and derived inventory. Deployers and future marketplace serving consume the resolved status pin instead of a moving branch or tag.

**What's here**
- `Plugin` v1alpha1 kind: spec = git/OCI source intent; status = server-resolved `ResolvedSource{Type,Commit,Digest}`, typed `Manifest`, derived `Inventory`, and readiness conditions.
- Plugin controller reconciliation that resolves and pins public git sources, classifies retryable versus terminal failures, and gates updates by `ObservedGeneration`.
- `AgentSource.Harness` for harness-based agents, with harness composition hoisted to top-level `AgentSpec` fields where appropriate: plugins, skills, MCP servers, model config, env, and permission mode.
- Deployment-facing `HarnessPolicy` API shape so AgentCore and other consumers can read the canonical harness composition from Agent specs.
- Plugin bundle/source/translate/materialize plumbing plus OpenAPI and generated UI client updates for the Plugin API and harness schema.

**Deferred follow-ups**
- Marketplace serving and arctl integration.
- OCI source resolution and optional registry-managed mirror/cache.
- Private git/authenticated source support.
- Additional deploy-time consumers beyond the AgentCore enterprise path.

# Change Type

/kind feature

# Changelog

```release-note
Add an experimental Plugin kind and harness-based agent source using source-pointer storage.
```

# Additional Notes

The branch intentionally keeps the source-pointer rationale in code and API behavior rather than carrying an ADR file in the PR. Phase 1 supports public git resolution; OCI and mirror durability remain follow-up work.


internal/cli/common/gitutil/gitutil.go
```diff
@@ -68,6 +81,14 @@ func ParseGitHubURL(rawURL string) (cloneURL, branch, subPath string, err error)
 // commit argument explicitly. branch is passed to `git clone --branch`; commit
 // triggers a fetch + checkout after the clone.
 func CloneAndCopy(repoURL, branch, commit, subPath, targetDir string, verbose bool) error {
```
lcopi: I don't think we need to keep this around, any caller that doesn't have an explicit context can pass context.Background directly

internal/cli/common/gitutil/gitutil.go
```diff
@@ -123,6 +152,123 @@ func CloneAndCopy(repoURL, branch, commit, subPath, targetDir string, verbose bo
...
+	}
+	return true
+}
+
+// ResolveRef resolves a branch, tag, or HEAD to a concrete commit SHA on the
+// remote WITHOUT cloning, using `git ls-remote`. A ref that is already a full
+// 40-char commit SHA is returned unchanged (lowercased). An empty ref (after
+// the URL-embedded branch is considered) resolves the remote's default branch
+// (HEAD). Only github.com URLs are supported (see ParseGitHubURL).
+func ResolveRef(repoURL, ref string) (string, error) {
```
lcopi: same as the above, this helper isn't really needed

.repotrail/harness-pr-stack-review.json
```diff
@@ -0,0 +1,1160 @@
+{
+  "repoTrailTour": 2,
+  "exportedAt": "2026-06-29T16:05:00Z",
```
lcopi: should this be committed?

internal/registry/controller/plugin_controller.go
```diff
@@ -0,0 +1,329 @@
...
+	"github.com/jackc/pgx/v5/pgxpool"
+	"k8s.io/client-go/util/workqueue"
+
+	"github.com/agentregistry-dev/agentregistry/internal/registry/plugins/bundle"
+	"github.com/agentregistry-dev/agentregistry/internal/registry/plugins/source"
+	"github.com/agentregistry-dev/agentregistry/pkg/api/v1alpha1"
+	pkgdb "github.com/agentregistry-dev/agentregistry/pkg/registry/database"
+	"github.com/agentregistry-dev/agentregistry/pkg/registry/v1alpha1store"
+)
+
```
lcopi, line 20: the plugin controller is busy and hiding lifecycle. I think a cleaner pattern would be to construct a controller, then call Start, which owns its goroutine and exposes Stop.

Something like:
```
type PluginController interface {
    Start(ctx context.Context) error
    Stop()
}

type pluginController struct {
    store    pluginStore
    resolver source.Resolver
    wakeups  <-chan struct{}
    resync   time.Duration
    queue    workqueue.TypedRateLimitingInterface[pluginQueueKey]

    cancel context.CancelFunc
    done   chan struct{}
}

func NewPluginController(cfg PluginControllerConfig) (PluginController, error) {
    if cfg.Store == nil {
        return nil, errors.New("plugin controller: store is required")
    }
    if cfg.Resolver == nil {
        return nil, errors.New("plugin controller: resolver is required")
    }
    if cfg.Resync == 0 {
        cfg.Resync = defaultControllerResyncInterval
    }

    return &pluginController{
        store:    cfg.Store,
        resolver: cfg.Resolver,
        wakeups:  cfg.Wakeups,
        resync:   cfg.Resync,
        queue:    newPluginQueue(),
        done:     make(chan struct{}),
    }, nil
}
```
Then Start explicitly owns the background lifecycle:
```
func (c *pluginController) Start(ctx context.Context) error {
    if c.cancel != nil {
        return errors.New("plugin controller: already started")
    }

    runCtx, cancel := context.WithCancel(ctx)
    c.cancel = cancel

    go func() {
        defer close(c.done)

        if err := c.run(runCtx); err != nil && !errors.Is(err, context.Canceled) {
            logger.Error("plugin controller stopped", "error", err)
        }
    }()

    return nil
}

func (c *pluginController) Stop() {
    if c.cancel != nil {
        c.cancel()
    }
    if c.done != nil {
        <-c.done
    }
}
```
And run is private:
```
func (c *pluginController) run(ctx context.Context) error {
    defer c.queue.ShutDown()

    workerErrs := make(chan error, 1)
    go func() { workerErrs <- c.runWorker(ctx) }()

    c.enqueueAllLogged(ctx)

    var ticks <-chan time.Time
    if c.resync > 0 {
        ticker := time.NewTicker(c.resync)
        defer ticker.Stop()
        ticks = ticker.C
    }

    for {
        select {
        case <-ctx.Done():
            return ctx.Err()
        case err := <-workerErrs:
            return err
        case <-c.wakeups:
            c.enqueueAllLogged(ctx)
        case <-ticks:
            c.enqueueAllLogged(ctx)
        }
    }
}
```
Then App(...) does the actual composition / orchestration
```
pluginController, err := controller.NewPluginController(controller.PluginControllerConfig{
    Store:    stores[v1alpha1.KindPlugin],
    Resolver: pluginsource.NewGitResolver(),
    Wakeups:  controller.ControlPlaneWakeups(ctx, pool), // or build this inside config helper
    Resync:   controller.DefaultControllerResyncInterval,
})
if err != nil {
    return fmt.Errorf("new plugin controller: %w", err)
}

if err := pluginController.Start(ctx); err != nil {
    return fmt.Errorf("start plugin controller: %w", err)
}
defer pluginController.Stop()
```

internal/registry/controller/skill_controller.go
```diff
@@ -0,0 +1,346 @@
...
+	Wakeups <-chan struct{}
+
+	queueMu sync.Mutex
+	queue   workqueue.TypedRateLimitingInterface[skillQueueKey]
+}
+
+// StartSkillController wires the Skill controller and starts it in the
+// background, opening its own control-plane LISTEN subscription (same mechanism
+// as the Plugin and Deployment controllers).
+func StartSkillController(
```
lcopi: it looks like there is a lot of crossover between this controller and the plugin controller, is there any way to share the controller generics and have it operate across all the types? If not I think this should be refactored to the same contructor/start/stop lifecycle interface pattern as the plugin controller

pkg/registry/v1alpha1store/migrations/010_plugins_table.up.sql
```diff
@@ -0,0 +1,38 @@
...
+    annotations jsonb DEFAULT '{}'::jsonb NOT NULL,
+    spec jsonb NOT NULL,
+    content_hash character(64) NOT NULL,
+    status jsonb DEFAULT '{}'::jsonb NOT NULL,
+    created_at timestamp with time zone DEFAULT now() NOT NULL,
+    updated_at timestamp with time zone DEFAULT now() NOT NULL,
+    deletion_timestamp timestamp with time zone,
+    PRIMARY KEY (namespace, name, tag)
+);
+
```
lcopi, line 23: For indexes I would recommend this set instead where the comment ties it to the current data lookup in the code:

```
// list by labels
  CREATE INDEX IF NOT EXISTS plugins_labels_gin
      ON plugins USING gin (labels);

// list by tags
  CREATE INDEX IF NOT EXISTS plugins_list_alive
      ON plugins USING btree (namespace, name, tag, updated_at)
      WHERE deletion_timestamp IS NULL;

// list tags for a single plugin
  CREATE INDEX IF NOT EXISTS plugins_tags_alive
      ON plugins USING btree (namespace, name, updated_at DESC, tag DESC)
      WHERE deletion_timestamp IS NULL;

// purge terminating
  CREATE INDEX IF NOT EXISTS plugins_terminating
      ON plugins USING btree (deletion_timestamp)
      WHERE deletion_timestamp IS NOT NULL;
```

currently in the code data is not searched by `plugins_updated_at_desc` or `plugins_spec_gin`

.repotrail/harness-pr-stack-review.json
ilackarms: Accidental commit, removed.

internal/cli/common/gitutil/gitutil.go
ilackarms: Done. Removed the context-free helper and updated callers to use the context-aware form with `context.Background()` where there is no explicit caller context.

internal/cli/common/gitutil/gitutil.go
ilackarms: Done. Removed this context-free wrapper as well; callers now use `ResolveRefContext` directly.

internal/registry/controller/plugin_controller.go
ilackarms: Done. Refactored the plugin controller into an explicit constructor plus `Start(ctx)` / `Stop()` lifecycle so goroutine ownership is visible at the composition root.

internal/registry/controller/skill_controller.go
ilackarms: I kept this minimal for this PR: the skill controller now follows the same constructor/start/stop lifecycle shape as the plugin controller, but I did not introduce a shared generic framework yet. That feels like a follow-up once the two controllers stabilize.

pkg/registry/v1alpha1store/migrations/010_plugins_table.up.sql
ilackarms: Done. Replaced the broad indexes with lookup-specific indexes for labels, live plugin listing, live tag listing, and terminating-row purge behavior, with matching down migration cleanup.

internal/cli/common/gitutil/gitutil.go
```diff
@@ -123,6 +149,119 @@ func CloneAndCopy(repoURL, branch, commit, subPath, targetDir string, verbose bo
...
+// positionally to git, never through a shell, so no further quoting is needed.
+func safeGitRef(ref string) error {
+	if strings.HasPrefix(ref, "-") {
+		return fmt.Errorf("invalid git ref %q: must not start with '-'", ref)
+	}
+	return nil
+}
+
+// isFullCommitSHA reports whether s is a full 40-character hex commit SHA.
+func isFullCommitSHA(s string) bool {
```
lcopi, line 164: nit: could use https://pkg.go.dev/github.com/go-git/go-git/v5@v5.19.1/plumbing#IsHash for this

internal/cli/common/gitutil/gitutil.go
```diff
@@ -123,6 +149,119 @@ func CloneAndCopy(repoURL, branch, commit, subPath, targetDir string, verbose bo
...
+		ref = urlBranch
+	}
+	lsRef := ref
+	if lsRef == "" {
+		lsRef = "HEAD"
+	}
+	if err := safeGitRef(lsRef); err != nil {
+		return "", err
+	}
+	out, err := exec.CommandContext(ctx, "git", "ls-remote", cloneURL, lsRef).Output()
```
lcopi, line 201: would it be better to use a github client for this instead of os.exec? using the client to get the sha https://github.com/google/go-github/blob/v84.0.0/github/github-accessors.go#L16258

internal/registry/plugins/bundle/bundle.go
```diff
@@ -0,0 +1,137 @@
...
+	MaxBundleBytes int64 = 128 << 20 // 128 MiB
+)
+
+// CanonicalBundle is the portable core of a plugin: a flat, path-keyed set of
+// files. It is NOT harness-specific; translation to a harness's on-disk layout
+// happens at deploy time.
+//
+// Paths are clean, relative, forward-slash separated (no leading "/" or "..").
+type CanonicalBundle struct {
+	Files map[string][]byte
```
lcopi, line 46: wondering if we should use a type for the files here, it would allow us to expand metadata in the future. For example

```
type File struct {
  Data []byte
  Mode fs.FileMode
}
```

not sure if we care about the permissions of the file or if we're controlling that where its consumed?

internal/registry/plugins/bundle/manifest.go
```diff
@@ -0,0 +1,178 @@
...
+// BuildInventory indexes a canonical bundle into a PluginInventory: the skills,
+// sub-agents, commands, MCP servers, hooks, and bin/ executables it actually
+// ships — the legible governance risk surface, derived by scanning bundle files
+// (not the author-supplied manifest). Best-effort: a malformed declarative file
+// is skipped rather than failing the resolve. Output is deterministic (sorted).
+func BuildInventory(b *CanonicalBundle) *v1alpha1.PluginInventory {
+	m := &v1alpha1.PluginInventory{}
+
+	var skillPaths, agentPaths, cmdPaths, binPaths []string
+	for p := range b.Files {
```
lcopi: nit: I think you could get rid of the intermediate sorts by ranging over the sorted key list:

```
for _, p := range slices.Sorted(maps.Keys(b.Files)) {
    switch {
    case isSkillPath(p):
        name, desc := parseSkillFrontmatter(b.Files[p])
        if name == "" { name = skillNameFromPath(p) }
        m.Skills = append(m.Skills, v1alpha1.PluginSkill{Name: name, Description: desc})
    case isAgentPath(p):
        m.Agents = append(m.Agents, baseNameNoExt(p))
    case isCommandPath(p):
        m.Commands = append(m.Commands, baseNameNoExt(p))
    case isBinPath(p):
        m.Executables = append(m.Executables, strings.TrimPrefix(p, "bin/"))
    }
}
if data, ok := b.Files[".mcp.json"]; ok { m.MCPServers = parseMCPServers(data) }
if data, ok := b.Files["hooks/hooks.json"]; ok { m.Hooks = parseHooks(data) }
```

pkg/api/v1alpha1/plugin_validate.go
```diff
@@ -0,0 +1,93 @@
...
+	} else {
+		for _, e := range validatePluginSource(s.Source) {
+			errs.Append("spec.source."+e.Path, e.Cause)
+		}
+	}
+	return errs
+}
+
+// isFullCommitSHA reports whether s is a full 40-character hex commit SHA.
+func isFullCommitSHA(s string) bool {
```
lcopi, line 34: it looks like a lot of this is shared with gitutil, can it use those validators instead?

internal/cli/common/gitutil/gitutil.go
ilackarms: Checked this. `plumbing.IsHash` would work, but this repo does not currently depend on go-git. Adding that dependency just for a 40-hex predicate feels heavier than the local helper, so I am going to keep this local for now.

internal/cli/common/gitutil/gitutil.go
ilackarms: I would keep `git ls-remote` here. It preserves Git semantics for branch/tag/default HEAD resolution, works without GitHub API auth or rate limits, and matches the rest of this utility, which already shells out for clone/fetch/checkout.

internal/registry/plugins/bundle/bundle.go
ilackarms: Reasonable future extension, especially if we decide to preserve executable modes or other file metadata. For this PR, nothing consumes mode from `CanonicalBundle`; materialization controls permissions later. We are going to revisit file bundling in more depth in a follow-up, so I would rather keep this simple here and design the richer file shape with the full bundling contract in view.

internal/registry/plugins/bundle/manifest.go
ilackarms: Done in b6eb204c. `BuildInventory` now ranges over `slices.Sorted(maps.Keys(b.Files))` and appends directly to the inventory buckets, keeping the same deterministic output with less bookkeeping.

pkg/api/v1alpha1/plugin_validate.go
ilackarms: I checked the layering here. The overlap is just the full-SHA predicate; `pkg/api/v1alpha1` should not import `internal/cli/common/gitutil`, and creating a new shared package solely for this helper does not feel better than the small local duplication. I would keep it local for now. If we add more shared git/ref validation behavior later, that would be the right point to extract a real shared package.

lcopi (approved):

xytian315 (approved):

Issue #432, opened 2026-04-15
Inconsistent approach towards logging


timflannagan:
Our approach to logging is very inconsistent across the CLI & server components. We use fmt.PrintX substantially, log.PrintX ~19 times in the server, and then pkg/logging from the upstream package in other places. For the registry server, we should at least standardize on the pkg/logging package and determine what we want to do with CLI logging. Enforcement via golangci-lint should be explored too.

timflannagan:
This has largely been solved. The only gap is with the legacy kubernetes runtime implementation.

PR #556, opened 2026-06-22
internal: Use logger for kubernetes applies

timflannagan:
# Description

Route verbose Kubernetes runtime apply messages through the shared logging package instead of writing directly to stdout. This keeps the runtime helper aligned with the server logging pipeline while leaving CLI-specific output conventions alone.

Refs #432.

# Change Type

/kind cleanup

# Changelog

```release-note
NONE
```

# Additional Notes

Verified with `go test ./internal/registry/runtimes/kubernetes`.

### DIFF
--- a/internal/registry/runtimes/kubernetes/platform.go
+++ b/internal/registry/runtimes/kubernetes/platform.go
@@ -26,6 +26,7 @@ import (
 	runtimetypes "github.com/agentregistry-dev/agentregistry/internal/registry/runtimes/types"
 	runtimeutils "github.com/agentregistry-dev/agentregistry/internal/registry/runtimes/utils"
 	"github.com/agentregistry-dev/agentregistry/pkg/api/v1alpha1"
+	"github.com/agentregistry-dev/agentregistry/pkg/logging"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


internal/registry/runtimes/kubernetes/platform.go
```diff
@@ -43,6 +44,7 @@ var (
 	kubernetesNewClientForConfig   = func(restConfig *rest.Config) (client.Client, error) {
 		return client.New(restConfig, client.Options{Scheme: kubernetesScheme})
 	}
+	kubernetesLogger = logging.New("kubernetes-runtime")
```
lcopi, line 47: Nit: klog would be easier to type than kubernetesLogger

lcopi (approved):

internal/registry/runtimes/kubernetes/platform.go
timflannagan: Yeah noticed the same thing while reviewing the code locally. It's not just the name of the variable too; we should add naming here, e.g. adapter/kubernetes, adapter/kagent, etc. Decided against that for now since there's there's cruft to burn down here where the first step is standardizing on this logging pattern, then making it better, etc.

timflannagan: Let's talk tomorrow (more context to come)

PR #574, opened 2026-07-10
test(db): fail instead of skip when Postgres is unreachable

nikolasmatt:
# Description

The integration-tagged DB test helpers (`v1alpha1store.NewTestPool` and the `freshDB`/`newDB` copies under `pkg/registry/database`) skip when Postgres is unreachable, so `make test` without a database goes silently green. They now fail fast with a hint to start Postgres or run `make test-unit`. Failure messages never echo the DSN (connect errors already name user/database/host, and credentials stay out of CI logs). CI already provisions Postgres before `make test` and is unaffected.

Also:

- The three scattered DB-helper test files (`legacymigrate`, `migrate_searchpath`, `orchestrator`) consolidate into one `test/integration/database/` package with a single shared `freshDB`; still covered by `make test`'s `-tags=integration ./...`.
- The admin DSN can be overridden via `AGENT_REGISTRY_TEST_DATABASE_URL` (URL-form; default unchanged) or passed explicitly via `NewTestPoolWithDSN(t, dsn) (pool, testDSN)` for suites that own their database configuration; the exported test surface is exactly `NewTestPool` + `NewTestPoolWithDSN`, with the DSN plumbing unexported behind them (pinned by in-package tests).
- The `test-cli-e2e` tier's documented self-skip is untouched.

# Change Type

/kind cleanup

# Changelog

```release-note
NONE
```


pkg/registry/database/dbtest/dbtest.go
```diff
@@ -0,0 +1,92 @@
...
+	"testing"
+
+	"github.com/jackc/pgx/v5"
+)
+
+// AdminDSN returns the admin connection URI for integration-test DB helpers:
+// AGENT_REGISTRY_TEST_DATABASE_URL when set, otherwise the local dev default
+// (localhost:5432, user/password agentregistry). The value must be a
+// URL-form DSN (postgres://...); keyword/value DSNs are not supported.
+func AdminDSN() string {
```
lcopi: looks like this isn't used directly anywhere outside of ConnectAdmin, does it need to be exported?

Seems a bit awkward to load the dsn and return it without doing validation, should validation be wrapped into this helper?

pkg/registry/v1alpha1store/testutil_test.go
```diff
@@ -0,0 +1,78 @@
+//go:build integration
+
+package dbtest
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
```
lcopi, line 10: a bit overkill to test a test helper, but no complaints if you leave it

pkg/registry/database/dbtest/dbtest.go
```diff
@@ -0,0 +1,92 @@
...
+		}
+	}
+	u.RawQuery = q.Encode()
+	return u.Redacted()
+}
+
+// DBURI returns adminURI with its database replaced by dbName. A dbname
+// query parameter is dropped — pgx would apply it after the path, silently
+// redirecting every per-test URI back to the override's database.
+func DBURI(adminURI, dbName string) (string, error) {
```
lcopi: I think it would be simpler to accept an optional name param and build the admin URI with the db name at connect time, than reparse and inject after the fact. Not sure if there is a usecase for having both the hardcoded adminURI + the db named specific one?

AGENTS.md
```diff
@@ -270,17 +270,17 @@ func TestAgentService_GetAgent(t *testing.T) {
...
 make test
 
 # Run with coverage
 make test-coverage
 
 # Run with HTML coverage report
 make test-coverage-report
-
-# Run integration tests
-make test-integration
```
lcopi, line 283: I think it makes sense to allow running integration tests independently still, we should leave this as a unique make target independent of the unit tests

lcopi (approved):
some nit comments

AGENTS.md
nikolasmatt: `make test` is the integration make target. `make test-integration` doesn't exist. so this is just correcting the AGENTS.md instruction to stop tripping agents. Alternatively i can fix the make targets but i decided to not do it as part of this PR (not a strong opinion), happy to update them if you think it's worth it.

pkg/registry/v1alpha1store/testutil_test.go
nikolasmatt: fair. i started this the `RedactDSN` function more generically, so when i switched it to be a test helper i just kept the tests.

pkg/registry/database/dbtest/dbtest.go
nikolasmatt: > looks like this isn't used directly anywhere outside of ConnectAdmin, does it need to be exported?

Downstream projects that add their own db migrations sources need to execute these tests as part of their suite. This lets the 2 suites coordinate.

> Seems a bit awkward to load the dsn and return it without doing validation, should validation be wrapped into this helper?

🤔 probably should validate. Though validation already in ConnectAdmin.

lcopi (approved):

PR #576, opened 2026-07-13
Fix MCP compatibility endpoint being affected by AuthNZ checks

inFocus7:
# Description

The MCP compatibility endpoint still goes through authorization. This means that without a token, a user hits a 401/403 (I forget which is for what authNZ fail). This isn't an issue here in OSS, but any project that extends this **will** hit the issue if they implement their own Auth providers which would fail at this point. This endpoint should be completely **public**.

This fixes it by adding a new `PublicSession` session implementation, similar to our `SystemSession` as well as a new `publicPaths` middleware that attaches this session to any public paths. The `PublicSession` can be caught by any authz hook (e.g. a server listing) to decide on what to allow this session to access.

**Note**: `publicPaths` != `skipPaths`; may be better devX to rename, but the core difference is:
- The pre-existing `skipPaths` is for paths skipping authentication **and** authorization, e.g. `/health`, `/ping which don't require any form of authorization for access.
- The new `publicPaths` is for paths that can skip authentication (because they are public) **and** will require authorization checks for access to resources. By attaching the `PublicSession` context, callers can decide on access level.

# Change Type

```
/kind fix
```

# Changelog

More of an API change for people using this as a library than user-facing, but .

```release-note
Add new `PublicSession` implementation and `publicPaths` middleware to allow API extensions to define public user authorization (e.g. allow anonymous users to view all servers via the MCP compatibility servers.)
```

# Additional Notes

Nothing to test in this because OSS already has a lenient authNZ, this is a fix for extensions of this.


pkg/registry/auth/auth.go
```diff
@@ -84,6 +97,15 @@ func AuthnMiddleware(authn AuthnProvider, options ...MiddlewareOption) func(ctx
 			return
 		}
 
+		// Public paths bypass credential authentication but carry a
+		// PublicSession so downstream authz hooks can scope anonymous access.
+		for _, prefix := range config.publicPrefixes {
```
lcopi, line 102: can we add a test for this? To verify the public session is used for public prefixed routes

inFocus7: Just pushed tests to verify session set based on path & middleware 🫡

lcopi (approved):

PR #585, opened 2026-07-20
feat(model): fingerprint referenced Models

ilackarms:
# Description

Include `Deployment.spec.modelRef` in the default resolved dependency set used for apply fingerprinting. A referenced Model's identity, generation, and spec now participate in the fingerprint while the existing MCPServer, Plugin, Skill, and Prompt dependency snapshots remain intact.

This is the narrow prerequisite for enterprise AgentCore runtime translation to use the selected Model without adopting the legacy adapter fingerprinter hook, which returns only a fingerprint string and would lose operator-visible dependency evidence.

# Change Type

/kind feature

# Changelog

```release-note
Deployment apply fingerprints now include changes to referenced Models.
```

# Additional Notes

Validation:

- `go test -race ./pkg/types`
- `make test-unit` (1,228 tests)
- `make lint`
- clean-commit `make verify`
- `git diff --check`

### DIFF
--- a/pkg/types/fingerprint.go
+++ b/pkg/types/fingerprint.go
@@ -235,17 +235,39 @@ func materialHash(parts ...json.RawMessage) string {
 
 func defaultApplyDependencies(ctx context.Context, in ApplyInput) ([]v1alpha1.Object, error) {
 	agent, ok := in.Target.(*v1alpha1.Agent)
-	if !ok || agent == nil {
-		return nil, nil
-	}
+	hasModelRef := in.Deployment != nil && in.Deployment.Spec.ModelRef != nil
+	hasAgentRefs := ok && agent != nil &&
+		(len(agent.Spec.MCPServers) > 0 || hasHarnessCompositionRefs(in.Deployment, agent))
 	if in.Getter == nil {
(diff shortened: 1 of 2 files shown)


pkg/types/fingerprint.go
```diff
@@ -235,17 +235,39 @@ func materialHash(parts ...json.RawMessage) string {
...
-		if len(agent.Spec.MCPServers) > 0 || hasHarnessCompositionRefs(in.Deployment, agent) {
-			return nil, fmt.Errorf("fingerprint: getter required to resolve Agent dependency refs")
+		if hasModelRef || hasAgentRefs {
+			return nil, fmt.Errorf("fingerprint: getter required to resolve dependency refs")
 		}
 		return nil, nil
 	}
-	deps := make([]v1alpha1.Object, 0, len(agent.Spec.MCPServers)+len(agent.Spec.Plugins)+len(agent.Spec.Skills)+1)
+	dependencyCapacity := 0
+	if hasModelRef {
```
lcopi, line 248: nit: the capacity count makes the code less readable and it looks like the only benefit is scoping allocations on the array correctly. Not sure its worth that very minimal perf benefit

lcopi (approved):

PR #602, opened 2026-07-31
Add read-only Claude Code marketplace.json compatibility endpoint

danielle-mustillo:
# Description

Adds a read-only `GET /plugin-marketplace/marketplace.json` endpoint that re-exposes AgentRegistry's `Plugin` resources in the Claude Code `marketplace.json` shape (code.claude.com/docs/en/plugin-marketplaces), so a bare URL to this endpoint can be registered directly with `claude plugin marketplace add`.

- **Motivation:** let AgentRegistry serve as a source for `claude plugin marketplace add` without a separate export step.
- **What changed:**
  - `pkg/pluginmarketplace`: translates a resolved `v1alpha1.Plugin` source pin into the `marketplace.json` `url`/`git-subdir` source forms (phase 1 — no `github` shorthand form, since AgentRegistry always resolves to a concrete repository URL).
  - `internal/registry/api/handlers/pluginmarketplace`: the HTTP handler, walking every Plugin row across all namespaces (paginated internally) and skipping any not-yet-resolved or unsupported-source entries so the document never contains a partial/broken entry. Honors the same optional per-kind RBAC `ListFilter` as the native Plugin read path.
  - `internal/registry/config` / router wiring: new `PLUGIN_MARKETPLACE_COMPAT_ENABLED` flag (default `false`) and optional path prefix, since the endpoint flattens every namespace into one unscoped catalogue.
  - `docs/plugin-marketplace-compatibility.md`: usage docs.

# Change Type

/kind feature

# Changelog

```release-note
Add optional read-only Claude Code marketplace.json compatibility endpoint (`PLUGIN_MARKETPLACE_COMPAT_ENABLED`), letting `claude plugin marketplace add` point directly at an AgentRegistry Plugin catalogue.
```

# Additional Notes

Off by default. The endpoint flattens every namespace into a single catalogue with no built-in scoping in OSS (nil `ListFilter`) — enable only where a full, unscoped Plugin catalogue is acceptable to expose (a public OSS registry, or behind a trusted gateway).


internal/registry/config/config.go
```diff
@@ -52,6 +52,24 @@ type Config struct {
...
+	// into one catalogue and honors the same per-kind RBAC list filter as the
+	// native Plugin read path (nil = public OSS behavior), so it is OFF by
+	// default — enable it only where a Plugin catalogue at this scope is
+	// acceptable (a public OSS registry, or behind a trusted gateway).
+	PluginMarketplaceCompatEnabled bool `env:"PLUGIN_MARKETPLACE_COMPAT_ENABLED" envDefault:"false"`
+	// PluginMarketplaceCompatPathPrefix optionally mounts the compatibility API
+	// under a base prefix (e.g. "/plugins"); empty serves the standard
+	// "/plugin-marketplace/marketplace.json" path at the root. Any prefix set
+	// here must match the base URL registered with the consuming agent.
+	PluginMarketplaceCompatPathPrefix string `env:"PLUGIN_MARKETPLACE_COMPAT_PATH_PREFIX" envDefault:""`
```
xytian315, line 70: question: should it be `AGENT_REGISTRY_PLUGIN_MARKETPLACE_*`

danielle-mustillo: ~~Its naming is aligned with the MCP_REGISTRY feature flags (scroll a bit up in this same file) which exposes a largely equivalent HTTP endpoint, so I propose to keep things consistent, if you agree?~~

danielle-mustillo: So I understand the concern now, this field doesn't match the design doc. That is by design though since line 112-116 has the following code:

```
err := env.ParseWithOptions(&cfg, env.Options{
		Prefix: "AGENT_REGISTRY_",
	})
```
So the net effect is that it has the AGENT_REGISTRY prefix as you noted. 

Hope this helps!

pkg/pluginmarketplace/translate.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+	if p.Status.ResolvedSource.Type != v1alpha1.PluginSourceTypeGit {
+		return PluginEntry{}, ErrUnsupportedSource
+	}
+	if p.Spec.Source == nil || p.Spec.Source.Git == nil || p.Spec.Source.Git.Repository == nil {
+		return PluginEntry{}, fmt.Errorf("plugin %q: resolved as git but has no git source spec", p.Metadata.Name)
+	}
+
+	repo := p.Spec.Source.Git.Repository
+	sha := p.Status.ResolvedSource.Commit
```
ilackarms, line 61: `FromPlugin` reads the repository URL/subfolder from `spec` but the pinned SHA from `status.resolvedSource` (which has no URL of its own). Since the controller keeps `Ready=True` while re-resolving a spec edit on an already-Ready plugin, there's a window where a just-edited plugin emits the new URL paired with the old repo's commit — a pin Claude Code can't install. Could we also require `status.observedGeneration >= metadata.generation` (mirroring `pluginReconciled`) and treat a mismatch as `ErrNotResolved`, so the entry is skipped until the new spec is actually resolved? A small test for the spec-edited-while-Ready case would lock this in.

internal/registry/api/handlers/pluginmarketplace/handlers.go
```diff
@@ -0,0 +1,163 @@
...
+	Body pluginmarketplace.MarketplaceResponse
+}
+
+func getMarketplace(cfg Config) func(context.Context, *struct{}) (*marketplaceOutput, error) {
+	return func(ctx context.Context, _ *struct{}) (*marketplaceOutput, error) {
+		name := cfg.MarketplaceName
+		if name == "" {
+			name = DefaultMarketplaceName
+		}
+		resp := pluginmarketplace.MarketplaceResponse{
```
ilackarms, line 101: With zero resolved plugins the document serializes as `"plugins": null` (the generated OpenAPI even reports `type: [array, "null"]`), while the marketplace.json schema expects an array — and an empty registry is exactly the state a first-time setup will be in when registering the URL. Could we initialize `Plugins` to an empty slice (the MCP compat endpoint does this via `make(..., 0, len(rows))`) so an empty catalogue emits `"plugins": []`? An empty-catalogue handler test would cover it.

internal/registry/api/handlers/pluginmarketplace/handlers.go
```diff
@@ -0,0 +1,163 @@
...
+					return nil, huma.Error500InternalServerError("decode plugin", err)
+				}
+				entry, err := pluginmarketplace.FromPlugin(p)
+				if err != nil {
+					// Not-yet-resolved / unsupported-source plugins are silently
+					// skipped: the marketplace.json document never contains a
+					// partial/broken entry.
+					continue
+				}
+				resp.Plugins = append(resp.Plugins, entry)
```
ilackarms, line 159: The catalogue flattens all namespaces but emits bare `metadata.name` as the marketplace entry name, so two same-named plugins in different namespaces yield duplicate `name` entries in one document — ambiguous for `claude plugin install <name>@<marketplace>`. Would you consider a collision policy here — e.g. skip-with-log on duplicates, or a namespace-qualified name (the MCP compat endpoint qualifies its names) — so multi-tenant catalogues stay unambiguous?

ilackarms (commented):
Nice, clean phase-1 cut — the reuse of the MCP compat pattern (flag-gated public path + per-kind ListFilter seam) makes the downstream RBAC story work with zero extra wiring, and the tests exercise the cursor walk properly. A few things around edge states of the served document below.

internal/registry/api/handlers/pluginmarketplace/handlers.go
```diff
@@ -0,0 +1,163 @@
...
+		Method:      http.MethodGet,
+		Path:        base + "/marketplace.json",
+		Summary:     "Get the plugin marketplace (Claude Code marketplace.json compatibility)",
+		Description: "Read-only listing of resolved plugins in the Claude Code marketplace.json format.",
+		Tags:        []string{"plugins"},
+	}, getMarketplace(cfg))
+}
+
+type marketplaceOutput struct {
+	Body pluginmarketplace.MarketplaceResponse
```
lcopi: nit: it's odd that the struct is unexported but the field is exported. What is this actually used for? can pluginmarketplace.MarketplaceResponse be used directly instead of creating this wrapper?

internal/registry/api/handlers/pluginmarketplace/handlers.go
```diff
@@ -0,0 +1,163 @@
...
+// doesn't set MarketplaceName.
+const DefaultMarketplaceName = "agentregistry"
+
+// pageLimit caps each internal List call while the handler walks the full
+// catalogue; it does not limit the size of the resulting document.
+const pageLimit = 100
+
+// PluginStore is the narrow read surface this handler needs from the Plugin
+// store. *v1alpha1store.Store satisfies it; tests supply a fake.
+type PluginStore interface {
```
lcopi, line 48: is this interface actually needed? seems like v1alpha1store.Store could be used directly?

internal/registry/api/handlers/pluginmarketplace/handlers_test.go
```diff
@@ -0,0 +1,213 @@
...
+	"github.com/agentregistry-dev/agentregistry/pkg/api/v1alpha1"
+	"github.com/agentregistry-dev/agentregistry/pkg/pluginmarketplace"
+	"github.com/agentregistry-dev/agentregistry/pkg/registry/resource"
+	"github.com/agentregistry-dev/agentregistry/pkg/registry/v1alpha1store"
+)
+
+// fakeStore is an in-memory PluginStore for handler tests. It paginates its
+// rows one at a time so tests exercise the handler's cursor-following loop,
+// and records every ListOpts it was called with.
+type fakeStore struct {
```
lcopi, line 27: Is there an existing fake that can be used instead of rolling a new one for this package?

pkg/pluginmarketplace/translate.go
danielle-mustillo: Very good catch. Agreed, I made the change as suggested. 

Its debatable that this is a cross cutting concern across many components, I did a bit of prototyping with this by adding a function in `status.go`, but it felt overkill to create a function there for this one usage. So I omitted that for now.

internal/registry/api/handlers/pluginmarketplace/handlers.go
danielle-mustillo: Also good catch. I confirmed that `null` will indeed cause Claude to reject adding the marketplace - which is obviously not acceptable. I made the change as requested with unit test, and the openapi spec was updated to only describe emitting a list.

internal/registry/api/handlers/pluginmarketplace/handlers.go
danielle-mustillo: I investigated a few different options, there are a lot of choices here, but ultimately I landed on adding the namespace-qualified name to the plugin name. I decided to trade off readability for no ambiguity and consistency. 

Format emitted is `namespace.name`. 

There are still very narrow situations where we can have name conflicts, namely when namespace + name combo results in the same characters. For example a plugin with namespace `test` name `p.lugin` and a second plugin with namespace `test.p` and name `lugin` would both emit `test.p.lugin` which would collide. This is a very narrow situation, arguably would never actually happen, but I added a first one wins logic in this case and a unit test for this situation. 

Regardless, all this to say, namespace-qualified names are now emitted.

internal/registry/api/handlers/pluginmarketplace/handlers.go
danielle-mustillo: Agreed.

Its not the only place in the codebase where this happens. See `mcpregistry/handlers.go` for example.

I found a way to resolve this by using an already existing and exported `Response` struct designed for this very purpose. The `huma.Register` function is still satisfied and there is code reuse.

internal/registry/api/handlers/pluginmarketplace/handlers.go
danielle-mustillo: I understand the concern but this is following the section "Interface Design" of the Claude.md file. Basically the claude.md instructs every major code module to depend on the smallest interfaces possible. `v1alpha1store.Store` has a lot of functionality that isn't needed in this case so depending on it here would tie us to those interfaces we don't use.

I can see it either way, I believe there is only one store anyway so an interface is a bit overkill at this point, but given this is not the only place we create this sort of interface (MCP does this too, and I'm sure there are others I haven't looked much), I think its better to align for now. 

Do you agree?

internal/registry/api/handlers/pluginmarketplace/handlers_test.go
danielle-mustillo: Unfortunately I don't think so, since every handler here needs to implement slightly different logic and this is a test stub. For example, this store lets us test with pagination (start, stop, cursor), which some other handlers don't support pagination at all. There is also an advantage of it not being exported out of this package the way it is implemented right now. I tend to agree with Claude in this one (mostly this is Claude's opinion ngl)

internal/registry/api/handlers/pluginmarketplace/handlers.go
lcopi: I think its a dumb rule in the claude.md :joy: 

But its fine to leave it in this pr

lcopi (approved):

PR #607, opened 2026-07-31
ci: Fix release artifact publication

timflannagan:
# Description

Clean up the release pipeline so stable releases have one supported entry
point: pushing a `vX.Y.Z` tag for a commit on `main`. The workflow validates
the tag before publication and builds both container images directly in GHCR
instead of targeting the local development registry.

Align the Helm chart with the published artifacts by retaining `X.Y.Z` as the
chart version while using `vX.Y.Z` for the application version and default
server image. Add a maintainer runbook covering the published artifacts,
tagging the freshly fetched `origin/main` revision, verification, and retry
behavior.

# Change Type

/kind cleanup

# Changelog

```release-note
NONE
```

# Additional Notes

Manual workflow dispatch is intentionally removed so the current release
process cannot publish an incomplete release without its Git tag and GitHub
Release. A future release-pipeline refactor can make dispatch the sole entry
point while having the workflow own tag creation.

Verified with `make charts-test CHART_VERSION=0.4.0` (105 tests),
`make charts-lint CHART_VERSION=0.4.0`, and `make release-cli VERSION=v0.4.0`.
All five CLI targets cross-compile successfully, the chart renders
`server:v0.4.0` with chart version `0.4.0` and application version `v0.4.0`,
and Docker build dry runs resolve both images under
`ghcr.io/agentregistry-dev/agentregistry`.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -4,17 +4,27 @@ on:
   push:
     tags:
     - "v*.*.*"
-  workflow_dispatch:
-    inputs:
-      version:
-        description: 'Version number'
 
 env:
-  VERSION: ${{ github.event.inputs.version || github.ref_name }}
+  VERSION: ${{ github.ref_name }}
   DOCKER_REGISTRY: ghcr.io
+  DOCKER_BUILD_REGISTRY: ghcr.io
 
 jobs:
+  validate:
+    name: Validate release version
+    runs-on: ubuntu-latest
+    steps:
+    - name: Validate version
+      shell: bash
+      run: |
+        if [[ ! "${VERSION}" =~ ^v(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)$ ]]; then
+          echo "::error::version must match vX.Y.Z; got ${VERSION}"
+          exit 1
+        fi
+
   docker:
+    needs: validate
     strategy:
       matrix:
         image:
@@ -121,7 +131,3 @@ jobs:
           bin/arctl-*
           build/charts/*.tgz
           build/charts/checksums.txt
-
-    - name: Log out of Helm registry
-      if: always()
-      run: $HELM registry logout "${HELM_REGISTRY}" || true
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -273,18 +273,9 @@ refactor(db): simplify query methods
 
 ## Building for Release
 
-```bash
-# Full clean build
-make all
-
-# Test the binary
-./bin/arctl version
-./bin/arctl ui
-
-# Create release
-git tag -a v1.0.0 -m "Release v1.0.0"
-git push origin v1.0.0
-```
+Maintainers should follow the complete [release process](docs/releasing.md).
+Releases are built and published by GitHub Actions from an annotated SemVer
+tag on `main`.
 
 ## Common Issues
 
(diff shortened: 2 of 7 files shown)


.github/workflows/release.yaml
```diff
@@ -121,7 +124,3 @@ jobs:
           bin/arctl-*
           build/charts/*.tgz
           build/charts/checksums.txt
-
-    - name: Log out of Helm registry
```
timflannagan, line 125: No need for this step for hosted GHA runners

.github/workflows/release.yaml
```diff
@@ -8,10 +8,13 @@ on:
     inputs:
       version:
         description: 'Version number'
+        required: true
+        type: string
 
 env:
   VERSION: ${{ github.event.inputs.version || github.ref_name }}
```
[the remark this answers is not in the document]
timflannagan: Addressed in 75a2acd. Manual dispatch has been removed; releases now run only from pushed stable vX.Y.Z tags, and validation gates artifact publication.

.github/workflows/release.yaml
```diff
@@ -4,17 +4,27 @@ on:
...
 
 jobs:
+  validate:
+    name: Validate release version
+    runs-on: ubuntu-latest
+    steps:
+    - name: Validate version
+      shell: bash
+      run: |
+        if [[ ! "${VERSION}" =~ ^v(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)$ ]]; then
```
timflannagan, line 21: Was tempted to emulate https://github.com/kgateway-dev/kgateway/blob/467617fbaeae1a4aa14a271657391a64bc77866a/.github/workflows/release.yaml#L100 but this is fine for now

lcopi (approved):

.github/workflows/release.yaml
```diff
@@ -4,17 +4,27 @@ on:
...
-  workflow_dispatch:
-    inputs:
-      version:
-        description: 'Version number'
 
 env:
-  VERSION: ${{ github.event.inputs.version || github.ref_name }}
+  VERSION: ${{ github.ref_name }}
   DOCKER_REGISTRY: ghcr.io
+  DOCKER_BUILD_REGISTRY: ghcr.io
```
timflannagan, line 11: Before I forget myself, needed as the Makefile itself splits up these variables. Should simplify later

timflannagan:
> ### 🟡 Not ready to approve
> CONTRIBUTING.md now describes releases as coming from an “annotated SemVer tag”, which is ambiguous vs the required `vX.Y.Z` tag format used elsewhere and could mislead maintainers.
> 
> _Once you've addressed the issues Copilot identified, you can request another Copilot review._
> 
> _This review doesn't count toward merge requirements. [Sign up for the private preview](https://forms.cloud.microsoft/r/zLCqnkB1FJ) to control whether Copilot approvals count._
> 
> Review details
> ### Suppressed comments (1)
> **CONTRIBUTING.md:278**
> 
> * This section says releases are published from an "annotated SemVer tag", but elsewhere in the docs and the release workflow the required tag format is explicitly `vX.Y.Z`. Calling it a SemVer tag can be read as `X.Y.Z` (no `v`), which would not trigger the release workflow and would conflict with the runbook.
> 
> ```
> Maintainers should follow the complete [release process](docs/releasing.md).
> Releases are built and published by GitHub Actions from an annotated SemVer
> tag on `main`.
> ```
> 
> * **Files reviewed:** 7/7 changed files
> * **Comments generated:** 0 new
> * **Review effort level:** Lite
> 
> We're testing this review assessment. Please use 👍 or 👎 to tell us if it's correct.

Insane

PR #606, opened 2026-07-31
cli: Remove daemon manager

timflannagan:
# Description

Remove the daemon manager and the `arctl daemon` command. AgentRegistry no longer embeds or controls a Docker Compose stack from the CLI; Kubernetes remains the default development environment, while local Docker operation uses the checked-in Compose file directly.

This removes the daemon command, its public manager and health-checking packages, the daemon-related types, and the daemon Makefile targets without compatibility shims. The local-development documentation now describes direct Compose lifecycle management.

# Change Type

/kind breaking_change

# Changelog

```release-note
Remove `arctl daemon`; manage AgentRegistry with Kubernetes or Docker Compose directly.
```

# Additional Notes

- `make run` continues to use Kubernetes.
- `make run-docker` and `make down` invoke `docker compose` directly.
- Verified with `make test-unit`.

PR #612, opened 2026-08-05
internal: Remove local deployment adapter

timflannagan:
# Description

The local Docker deployment adapter was built for demos and directly managed
host Docker Compose state from the registry process. That coupling is unsafe to
present as a supported runtime and does not match the current Kubernetes-only
deployment model.

This continues the cleanup from #605, which made E2E coverage Kubernetes-only,
#606, which removed the CLI daemon manager, and #609, which repaired the
Kubernetes deployment suite. The remaining local adapter left an untested
standalone path that looked production-ready even though it was unsupported.

This removes the adapter and its materialization code, the `Local` runtime type
and seed data, Docker socket and runtime configuration, UI deployment paths,
and now-unused dependencies and generated API fields. It also removes the
custom `arctl-agentgateway` image and release wiring, which existed only to
support the local adapter. Kubernetes is now the only built-in deployment
environment.

Existing databases receive a reversible forward migration that tombstones
active `Local` runtime rows. Fresh and upgraded installations therefore expose
the same supported runtimes without destroying state needed by a rollback. The
server image also drops its now-unused Docker CLI and Compose plugin.

The integration and E2E fixtures now use the migration-seeded Kubernetes
runtime instead of reconstructing demo-era local runtime state. Kubernetes
materialization helpers now live with their sole owning runtime rather than in
a shared utility package.

# Change Type

/kind breaking_change

# Changelog

```release-note
Remove the demo-only local Docker deployment runtime; Kubernetes is now the only built-in deployment environment.
```

# Additional Notes

The local `arctl run` developer workflow is unaffected; this change removes
only server-managed local deployments.

The migration-seeded `kubernetes-default` Runtime is temporary compatibility
behavior, not a long-term bootstrap contract. Follow-up work will stop seeding
runtime instances by default, make environment setup configure them
explicitly, and replace the current generic Kubernetes materialization path
with a direct kagent runtime integration.

The Makefile still provisions `kind-registry` as a Docker container on the
host. Replacing that with the native Kubernetes registry integration is worth a
separate follow-up and is intentionally outside this PR.

Verified with `make verify`, `make lint`, `go test ./...`, a targeted Postgres
upgrade and rollback test, Dockerfile validation, the affected
PostgreSQL-backed controller and MCP integration suites, an E2E package
compile, and the two previously failing batch-apply E2E tests against Kind.


internal/registry/runtimes/kubernetes/materialization.go
timflannagan, line 1: Ignore the exported functions/methods here. This will be fixed properly in a follow-up PR but didn't want to bloat an already large PR. Just wanted to localize here if the only consumer of the utils package was this adapter.

pkg/registry/v1alpha1store/migrations/013_remove_local_runtimes.up.sql
timflannagan, line 1: Note, didn't want to update the 001_ migration script which seeds the local and kubernetes-default adapters as it looks like golang-migrate doesn't do checksum properly? Maybe something to add as a CI check, not entirely sure yet.

The only thing to note (and you'll see this in a brittle test I had to change) is that this produces an event for the control_plane_events table we establish in the 009_ migration scripts to aid the deployment controller.

Overall, don't understand why any of these runtimes were seeded by default in the migration scripts. That seems like 1. an anti-pattern and 2. if we do want this seed logic, do it in the app via store upserts not via database migrations.

Because of this, also tempted to update this migration script to additionally drop the kubernetes-default seed too.

pkg/registry/v1alpha1store/migrations_upgrade_test.go
timflannagan, line 1: Don't _really_ think this test is necessary. Might rip this out, was just a sanity check from AI. I guess we could for now and rip later?

pkg/registry/v1alpha1store/store_test.go
```diff
@@ -798,6 +801,12 @@ func TestControlPlaneEventStore_PruneBeforeHonorsKeepAfterRevision(t *testing.T)
 	events := NewControlPlaneEventStore(pool, TestSchema())
 	ctx := context.Background()
 
+	// Migrations may legitimately emit control-plane invalidations. This test
+	// owns its event fixture, so discard bootstrap history before asserting
+	// exact prune counts.
+	_, err := pool.Exec(ctx, `TRUNCATE `+events.qualified+` RESTART IDENTITY`)
+	require.NoError(t, err)
```
timflannagan, line 808: Not sure if there's a smell here with our deployment controller design just yet. Brittle test for sure though

shashankram (approved):

internal/registry/controller/deployment_test.go
```diff
@@ -49,7 +49,7 @@ func deploymentFixture(desiredState string) *v1alpha1.Deployment {
 		},
 		Spec: v1alpha1.DeploymentSpec{
 			TargetRef:    v1alpha1.ResourceRef{Kind: v1alpha1.KindMCPServer, Name: "weather", Tag: "stable"},
-			RuntimeRef:   v1alpha1.ResourceRef{Kind: v1alpha1.KindRuntime, Name: "local"},
+			RuntimeRef:   v1alpha1.ResourceRef{Kind: v1alpha1.KindRuntime, Name: "kubernetes-default"},
```
lcopi, line 52: nit: could use the constant

internal/registry/controller/reconciler_integration_test.go
```diff
@@ -296,7 +288,7 @@ func TestDeploymentController_DeleteWaitsForRemoveThenPurgesFinalizedRow(t *test
 		Metadata: v1alpha1.ObjectMeta{Namespace: "default", Name: deployment.Metadata.Name},
 		Spec: v1alpha1.DeploymentSpec{
 			TargetRef:  v1alpha1.ResourceRef{Kind: v1alpha1.KindMCPServer, Name: "weather", Tag: v1alpha1store.DefaultTag()},
-			RuntimeRef: v1alpha1.ResourceRef{Kind: v1alpha1.KindRuntime, Name: "local"},
+			RuntimeRef: v1alpha1.ResourceRef{Kind: v1alpha1.KindRuntime, Name: "kubernetes-default"},
```
lcopi, line 291: nit for most of these might as well use the constant that was added

pkg/registry/v1alpha1store/migrations_upgrade_test.go
lcopi: agreed seems pointless

PR #625, opened 2026-08-18
add credentialref hook for private git repos

lcopi:
# Description

- **Motivation:** support private git repositories
- **What changed:** make git resolution flow through a credential hook 

# Change Type

```
/kind feature
```

# Changelog

```release-note
Support private git repos via a credential hook for downstream consumers
```

# Additional Notes


pkg/api/v1alpha1/common.go
```diff
@@ -11,4 +11,8 @@ type Repository struct {
 	Branch    string `json:"branch,omitempty" yaml:"branch,omitempty"`
 	Commit    string `json:"commit,omitempty" yaml:"commit,omitempty"`
 	Subfolder string `json:"subfolder,omitempty" yaml:"subfolder,omitempty"`
+
+	// SecretRef names credentials for a private repository. Key is ignored: git
+	// needs two values, so resolvers read a username/password pair.
+	SecretRef *SecretKeyRef `json:"secretRef,omitempty" yaml:"secretRef,omitempty"`
```
timflannagan: I don't think we should re-use SecretKeyRef here. As an aside, I don't think that struct should exist in the first place. 

I'd prefer something like this:

```go
CredentialsRef *LocalSecretReference `json:"credentialsRef,omitempty`
```

Since we aren't indexing into any secret key at all in this logic. WDYT?

internal/cli/common/gitutil/gitutil.go
```diff
@@ -219,7 +240,9 @@ func isFullCommitSHA(s string) bool {
 // (after the URL-embedded branch is considered) resolves the remote's default
 // branch (HEAD). ctx bounds the ls-remote call. A ref that resolves to no
 // commit returns ErrRefNotFound (terminal).
-func ResolveRefContext(ctx context.Context, repoURL, ref string) (string, error) {
+//
+// A non-nil auth authenticates against a private remote.
+func ResolveRefContext(ctx context.Context, repoURL, ref string, auth *url.Userinfo) (string, error) {
```
timflannagan, line 245: Looks like an existing problem, but I don't think ResolveRefContext needs to be exported?

internal/cli/common/gitutil/gitutil.go
```diff
@@ -122,10 +122,12 @@ func parseGitLabStyleURL(u *url.URL, parts []string, marker int) (cloneURL, bran
 // commit argument explicitly. branch is passed to `git clone --branch`; commit
 // triggers a fetch + checkout after the clone.
 //
+// A non-nil auth is spliced into the clone URL for a private repository.
+//
 // Every git invocation runs under ctx, so a caller can bound
 // clone/fetch/checkout time (and disk/CPU runaway) by passing a
 // context.WithTimeout. ctx cancellation kills the git child process.
-func CloneAndCopyContext(ctx context.Context, repoURL, branch, commit, subPath, targetDir string, verbose bool) error {
+func CloneAndCopyContext(ctx context.Context, repoURL, branch, commit, subPath, targetDir string, verbose bool, auth *url.Userinfo) error {
```
timflannagan, line 130: My god this entire package is pure spaghetti 😆.

internal/cli/common/gitutil/gitutil.go
timflannagan: Oh wow, there's a lot of cruft here. Nevermind, we'll fix in a refactor.

internal/cli/common/gitutil/source.go
```diff
@@ -0,0 +1,84 @@
...
+// Source pins and fetches git sources on behalf of the registry, resolving
+// per-repository credentials through a single hook. One Source is built at the
+// composition root and shared by every consumer (the Plugin and Skill
+// controllers today), so credentials enter the system in exactly one place.
+type Source struct {
+	credentials types.GitCredentialFunc
+}
+
+// NewSource returns a Source that authenticates through credentials. A nil hook
+// (the OSS default) fetches anonymously, so private repositories fail.
```
timflannagan, line 21: Just a temporary problem as far as i'm concerned. We should have proper private auth support soon here.

timflannagan (approved):
LGTM. We already talked; there's so much cruft in this area, we need to refactor it, and there's some testing gaps. I stopped making review comments midway as a result 😆. This will help unblock quite a bit of work.

Issue #627, opened 2026-08-20
Git-backed resources fail in the server image


timflannagan:
Git-backed Plugin reconciliation and branch/tag Skill resolution spawn system `git` through `internal/cli/common/gitutil`. The production server image installs only `curl`, so these paths fail with "executable file not found" and retry indefinitely as `SourceUnresolvable` while health checks remain green.

Replace server-side subprocess use with an embedded Go implementation such as go-git, or otherwise make the runtime dependency explicit. Add container E2E coverage that applies Git-backed Plugin and Skill resources and waits for `Ready=True`.

ibalat:
Reproduced on v0.4.0 (`aef6563`) on a k3s deployment of `ghcr.io/agentregistry-dev/agentregistry/server:v0.4.0`, with a concrete `Skill` rather than a Plugin — adding a data point since the issue mentions the missing container e2e coverage.

`kubectl exec` into the server container:

```
$ kubectl -n agentregistry exec deploy/agentregistry -- which git
NO_GIT
sh: 1: git: not found
```

A `Skill` whose `spec.source.repository.url` points at a reachable remote gets past `ParseGitURL` and fails at the `exec.CommandContext(ctx, "git", "ls-remote", …)` in `gitutil.ResolveRefContext`:

```yaml
status:
  conditions:
  - type: Ready
    status: "False"
    reason: SourceUnresolvable
    message: 'git ls-remote https://<host>/scm/<proj>/<repo>.git "master":
              exec: "git": executable file not found in $PATH'
```

Two observations that may be useful:

1. The pod stays `1/1 Running` throughout, matching the "health checks report green" note — nothing surfaces the missing runtime dependency except the per-resource condition.
2. `arctl pull` is unaffected, since it clones client-side from `Spec.Source.Repository.URL`. So the practical blast radius is `status.resolvedSource` (the commit pin) and anything that materializes from it at deploy time — a registry can still serve skills to `arctl pull` users while every Skill sits at `Ready=False`.

A `exec.LookPath("git")` check at controller start-up, surfaced as a startup error or a non-ready health check, would have turned this from a per-resource condition into an immediately visible misconfiguration.

timflannagan:
Agreed - more context is all this code is very fragile, especially since it relies on shelling out to git at server runtime. We're going to look at refactoring to git-go soon. In the meantime I can accept the 632 PR since it increases test coverage in this area, even if it'll be changed in the immediate term.

PR #632, opened 2026-08-24
fix(gitutil): accept plain clone URLs on any git host

### LINKED ISSUE #631: ParseGitURL rejects Bitbucket Server and every self-hosted host that is not github.com or *gitlab*
---

### Problem

`gitutil.ParseGitURL` gates on a hardcoded host allowlist:

```go
if gitLabMarker := slices.Index(parts, "-"); gitLabMarker >= 2 {
	return parseGitLabStyleURL(u, parts, gitLabMarker)
}
if strings.Contains(strings.ToLower(u.Host), "gitlab") {
	return parseGitLabRootURL(u, parts)
}
if u.Host != "github.com" {
	return "", "", "", fmt.Errorf("%w: %q", ErrUnsupportedHost, u.Host)
}
return parseGitHubStyleURL(u, parts)
```

Anything that is not `github.com` exactly, or does not contain the substring `gitlab`, is rejected — including Bitbucket Server / Data Center, Gitea, and GitHub Enterprise on a company domain. For an organisation whose only git hosting is Bitbucket Server there is no way to publish a skill at all: `ssh://` is rejected up front by the scheme check, so there is no fallback either.

Measured on v0.4.0 (`aef6563`) by applying a Skill per host and running `arctl pull skill`:

| `spec.source.repository.url` host | result |
|---|---|
| `github.com` | parsed, clone attempted |
| `gitlab.com` | parsed, clone attempted |
| `gitlab.int.example.com` | parsed, clone attempted |
| `mygitlab.corp.com` | parsed, clone attempted |
| `corp-gitlab.com` | parsed, clone attempted |
| `github.corp.com` | `parse Git URL: unsupported git host` |
| `bitbucket.org` | `parse Git URL: unsupported git host` |
| `bitbucket.int.example.com` | `parse Git URL: unsupported git host` |
| `git.example.com` | `parse Git URL: unsupported git host` |

The same parser backs the server-side controller, so the resource never leaves `Ready=False`:

```
reason: SourceUnsupported
message: 'parse Git URL: unsupported git host: "bitbucket.int.example.com"'
```

### The allowlist is already bypassable, which suggests it is not load-bearing

The GitLab `/-/` marker branch runs *before* the host check, and `parseGitLabStyleURL` reconstructs the clone URL by trimming and re-appending `.git`. So a GitLab-shaped path against a Bitbucket host parses fine and yields a correct Bitbucket clone URL:

```yaml
url: https://bitbucket.example.com/scm/proj/repo.git/-/tree/master/skills/my-skill
```

→ clone URL `https://bitbucket.example.com/scm/proj/repo.git`, branch `master`, subPath `skills/my-skill`. Verified end to end against a real Bitbucket Server instance: `arctl pull skill` clones and copies the subfolder correctly.

That is the only workaround available today, and it depends on an undocumented parse-order detail. It also shows the allowlist is not protecting anything — it is only selecting a *path shape*, and it is easy to opt out of.

### Suggested fix

`Repository` already carries `branch`, `commit` and `subfolder`, so a plain clone URL needs no path-shape parsing at all:

```go
type Repository struct {
	URL       string
	Branch    string
	Commit    string
	Subfolder string
}
```

Proposal: keep the GitHub/GitLab path parsers as a convenience for pasted browser URLs, but when the URL is already a plain clone URL (path ends in `.git`), accept it for any host and take branch/subfolder from the explicit fields. Optionally add a `provider` hint on `Repository` for the ambiguous cases. This would make Bitbucket, Gitea and self-hosted GitHub work without a per-provider parser each.

### Environment

- arctl `v0.4.0`, git commit `aef6563`, build date 2026-08-03
- server `v0.4.0` (same commit)
- host allowlist confirmed unchanged on current `main`

### Related

- #265 / #277 — GitLab publish-source naming; `--github` renamed to `--git`, host parsing untouched
- #548 — GitLab support for skill publish/pull
- #627 — separate issue: the server image has no `git` binary, so even a supported host cannot resolve

ibalat:
# Description

**Motivation:** `gitutil.ParseGitURL` gates on a host allowlist — `github.com` exactly, or any host containing the substring `gitlab` — so Bitbucket Server / Data Center, Gitea, and GitHub Enterprise on a company domain are rejected with `ErrUnsupportedHost`. `ssh://` is rejected by the scheme check as well, so an organisation whose only git hosting is Bitbucket has no way to publish a skill at all. The same parser backs the server-side controllers, so the resource never leaves `Ready=False` / `SourceUnsupported`.

**What changed:** the allowlist only ever selected a *path shape*. The GitHub and GitLab parsers exist to infer a branch and subdirectory from a web URL's layout; a plain clone URL has no layout to infer, and `Repository` already carries `Branch`, `Commit` and `Subfolder` explicitly. So an unknown host needs no parser of its own — only its clone URL:

- a path whose last segment ends in `.git` is accepted on **any** host and reduced to the repository root
- the GitHub and GitLab parsers are untouched, so pasted browser URLs keep working
- a non-clone URL on an unknown host still returns `ErrUnsupportedHost`, now with a message pointing at the clone URL
- `parseGitLabRootURL` → `parseRepoRootURL`: a GitLab project root and a plain clone URL reduce to the same shape, so the two paths share one host-agnostic helper

Worth noting for reviewers: the allowlist was already bypassable, which is part of why it looks non-load-bearing. The GitLab `/-/` marker branch runs *before* the host check, so `https://bitbucket.example.com/scm/proj/repo.git/-/tree/master/skills/my-skill` parses today and yields a correct Bitbucket clone URL. That undocumented parse-order detail is the only workaround available on v0.4.0. This PR makes the supported spelling — the plain clone URL — work instead.

**Related issues:** Fixes #631

# Change Type

/kind fix

# Changelog

```release-note
Git sources on hosts other than GitHub and GitLab (Bitbucket Server, Bitbucket Cloud, Gitea, GitHub Enterprise on a company domain) are now accepted when the URL is the repository clone URL ending in `.git`; branch and subfolder come from the explicit `Repository` fields.
```

# Additional Notes

**Verified end to end against a real Bitbucket Server instance.** Same registry, same `Skill`, `url: https://<host>/scm/<proj>/<repo>.git` with `branch: master` and `subfolder: skills/vm-to-pod`:

```
--- released arctl v0.4.0 ---
Cloning https://<host>/scm/<proj>/<repo>.git (branch master) into /tmp/p_old
Error: parse Git URL: unsupported git host: "<host>"

--- this branch ---
(subfolder hint: skills/vm-to-pod)
Pulled fixtest
SKILL.md  assets  references  skill.yaml
```

**Tests:** eight cases added to the `TestParseGitURL` table — Bitbucket Server (with and without a trailing slash), Bitbucket Cloud, Gitea, GitHub Enterprise on a company domain, plus two rejections (a Bitbucket Server `/projects/…/browse` web URL and an unknown-host repo root without `.git`). `TestParseGitURLUnsupportedHost` pins the sentinel, because `skill_controller.go` and `plugins/source` classify on `errors.Is(err, ErrUnsupportedHost)` to decide the failure is terminal.

**Not addressed here:** #627 (the server image ships no `git` binary) blocks `Ready=True` for every git-backed resource regardless of host, so a Bitbucket source still will not reconcile server-side until that is fixed. Client-side `arctl pull` is unaffected and works with this change.

**Scope choice:** I kept `Repository.provider` out of it. The `.git` suffix is a sufficient and unambiguous signal on every git server, and adding an API field would be a larger surface change than the bug needs. Happy to add it if you would rather have the explicit hint.

### DIFF
--- a/internal/cli/common/gitutil/gitutil.go
+++ b/internal/cli/common/gitutil/gitutil.go
@@ -16,9 +16,11 @@ import (
 )
 
 var (
-	// ErrUnsupportedHost is returned for a non-GitHub/non-GitLab host. It is a
-	// permanent condition: callers can errors.Is it to avoid retrying hosts that
-	// are not supported.
+	// ErrUnsupportedHost is returned for a web URL on a host whose layout is not
+	// understood — every host but github.com and GitLab. Such a host is still
+	// usable through its plain clone URL (see ParseGitURL). It is a permanent
+	// condition: callers can errors.Is it to avoid retrying a URL that cannot
+	// be parsed.
 	ErrUnsupportedHost = errors.New("unsupported git host")
 	// ErrRefNotFound is returned when a ref resolves to no commit on the remote
 	// (deleted branch/tag, typo, or a short/non-existent SHA). Terminal:
@@ -32,6 +34,13 @@ var (
 // Branch names containing slashes (e.g. feature/my-branch) are supported when
 // encoded as %2F in the URL. The raw (escaped) path is used for splitting so
 // the encoded branch segment is preserved, then unescaped for the return value.
+//
+// Any other host — Bitbucket Server, Gitea, GitHub Enterprise on a company
+// domain — is accepted through its plain clone URL, the path ending in ".git".
+// There is no web-URL layout to infer a branch or subdirectory from in that
+// case, so both come back empty and the caller supplies them explicitly
+// (Repository.Branch and Repository.Subfolder). A non-clone URL on such a host
+// returns ErrUnsupportedHost.
 func ParseGitURL(rawURL string) (cloneURL, branch, subPath string, err error) {
 	u, err := url.Parse(rawURL)
 	if err != nil {
@@ -60,13 +69,19 @@ func ParseGitURL(rawURL string) (cloneURL, branch, subPath string, err error) {
 		return parseGitLabStyleURL(u, parts, gitLabMarker)
 	}
 	if strings.Contains(strings.ToLower(u.Host), "gitlab") {
-		return parseGitLabRootURL(u, parts)
+		return parseRepoRootURL(u, parts)
 	}
-	if u.Host != "github.com" {
-		return "", "", "", fmt.Errorf("%w: %q", ErrUnsupportedHost, u.Host)
+	if u.Host == "github.com" {
+		return parseGitHubStyleURL(u, parts)
+	}
+	// Unknown host: accept it only as a plain clone URL, which is unambiguous
+	// on every git server. Guessing a web layout instead would silently
+	// mis-parse namespaces (Bitbucket Server's /scm/ prefix, Gitea's /raw/).
+	if strings.HasSuffix(parts[len(parts)-1], ".git") {
+		return parseRepoRootURL(u, parts)
 	}
 
-	return parseGitHubStyleURL(u, parts)
+	return "", "", "", fmt.Errorf("%w: %q (pass the repository clone URL, ending in .git)", ErrUnsupportedHost, u.Host)
 }
 
 func parseGitHubStyleURL(u *url.URL, parts []string) (cloneURL, branch, subPath string, err error) {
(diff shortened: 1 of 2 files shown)


ibalat:
@timflannagan hi Tim, I fixed the lint, please can you review it again?

timflannagan (approved):

PR #638, opened 2026-08-25
internal: Add label filter and display to arctl get

timflannagan:
# Description

`arctl get` had no way to filter or display resource labels, even though
content-registry kinds (agents, mcps, skills, prompts, plugins, models)
already carry a queryable `labels` map on their metadata.

This adds two flags:

- `-l/--labels`: filters list output by comma-separated `key=value` pairs,
  forwarded through `scheme.ListOpts` into the existing
  `client.ListOpts.Labels` query param. It's restricted to content kinds
  (those with `ListTags` set) and to list mode, mirroring the existing
  `--tag`/`--latest` guards. It's rejected under `get all` for now since
  that command builds a fresh `ListOpts` per kind and mixes in kinds that
  don't support the filter at all — extending it there is a natural
  follow-up.
- `--show-labels`: a kubectl-style flag that appends a trailing `LABELS`
  column to table output with each row's sorted `key=value` pairs (or
  `<none>`). It reads metadata directly off the `v1alpha1.Object`
  interface, so it works uniformly across every registered kind without
  touching per-kind `RowFunc` implementations.

# Change Type

/kind feature

# Changelog

```release-note
Add `-l/--labels` and `--show-labels` flags to `arctl get` for filtering and
displaying resource labels.
```

# Additional Notes

Follow-ups intentionally left out of this PR:

- `arctl get all -l <label>` selection support. `runGetAll` currently builds
  a fresh `ListOpts` per kind itself rather than threading the caller's
  flags through, and mixes in non-content kinds that don't support the
  filter at all — extending it needs a small refactor of that dispatch path.
- `arctl delete -l <label>` (bulk delete by label selector). `delete` has no
  list mode today — it requires an explicit `TYPE NAME` or `-f FILE` — so
  this would be a new list-then-delete flow with its own safety
  considerations (partial-failure handling, possibly a confirmation or
  `--dry-run`), not a small addition.
- Revisiting whether `--all-tags` should be reworked or removed, possibly in
  the same pass as the `get all -l` refactor above.
- Label selection intentionally stays simple key=value equality matching,
  not kubectl's fuller `--selector` expression language (set-based
  `in`/`notin`/exists operators). The current v1alpha1 metadata envelope
  doesn't need that expressiveness yet.
- `internal/cli/declarative` could use some general cleanup/TLC independent
  of this change.

### DIFF
--- a/internal/cli/declarative/get.go
+++ b/internal/cli/declarative/get.go
@@ -4,7 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"sort"
+	"slices"
 	"strings"
 
 	"github.com/spf13/cobra"
@@ -369,7 +369,7 @@ func formatLabels(item any) string {
 	for key := range meta.Labels {
 		keys = append(keys, key)
 	}
-	sort.Strings(keys)
+	slices.Sort(keys)
 	pairs := make([]string, len(keys))
 	for i, key := range keys {
 		pairs[i] = key + "=" + meta.Labels[key]
--- a/internal/cli/declarative/get.go
+++ b/internal/cli/declarative/get.go
@@ -52,7 +52,7 @@ Examples:
 	}
 	cmd.Flags().StringP("output", "o", "table", "Output format: table, yaml, json")
 	cmd.Flags().StringVarP(&labels, "labels", "l", "", "Content kinds only: filter list rows by comma-separated key=value labels.")
-	cmd.Flags().Bool("show-labels", false, "Table output only: print an additional LABELS column with each resource's labels.")
+	cmd.Flags().Bool("show-labels", false, "Print an additional LABELS column with each resource's labels; ignored for -o yaml/json, which already include labels.")
 	cmd.Flags().String("tag", "", "Tagged kinds only. With NAME: fetch one tag (defaults to latest). Without NAME: filter the list to this tag.")
 	cmd.Flags().Bool("latest", false, "List mode only: restrict to rows pinned to the literal 'latest' tag (equivalent to --tag latest).")
 	cmd.Flags().Bool("all-tags", false, "List every tag of NAME (tagged content kinds only)")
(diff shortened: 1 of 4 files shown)


internal/cli/declarative/get.go
```diff
@@ -59,6 +65,7 @@ func runGet(cmd *cobra.Command, deps cliruntime.Deps, args []string) error {
 	outputFormat, _ := cmd.Flags().GetString("output")
 	allTags, _ := cmd.Flags().GetBool("all-tags")
```
timflannagan, line 66: The --all-tags field is awkward to me, might investigate ripping it out as a follow up, along with arctl get all -l selection support, which would require some refactors to the current code that I wanted to hold off on for now.

internal/cli/declarative/get.go
```diff
@@ -47,6 +51,8 @@ Examples:
 		},
 	}
 	cmd.Flags().StringP("output", "o", "table", "Output format: table, yaml, json")
+	cmd.Flags().StringVarP(&labels, "labels", "l", "", "Content kinds only: filter list rows by comma-separated key=value labels.")
```
timflannagan, line 54: I think k/v label selection in sufficient for now given our v1alpha1 envelope, didn't bring in the --selector kubectl approach that allows you construct more advanced selection.

internal/cli/declarative/get.go
timflannagan, line 1: Some of this package needs a little TLC at some point too

internal/cli/declarative/get.go
```diff
@@ -47,6 +51,8 @@ Examples:
...
+	cmd.Flags().Bool("show-labels", false, "Table output only: print an additional LABELS column with each resource's labels.")
```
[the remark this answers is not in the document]
timflannagan: Reworded the flag help text to reflect the actual behavior (silently ignored for -o yaml/json, matching kubectl's --show-labels) instead of implying enforcement. See 2f92239.

lcopi (approved):

PR #683, opened 2026-09-17
resource: extract the ObjectStore contract behind the generic handlers

ilackarms:
# Description

The generic resource pipeline (`resource.Register`, the batch `/v0/apply` endpoint, and the production admission functions) hard-codes `*v1alpha1store.Store`. An extension kind whose persistence needs a different transaction boundary, for example one that authorizes inside the store and commits its own writes, cannot be served through the shared handlers today and has to duplicate them.

This change extracts the store contract the pipeline actually drives into `resource.ObjectStore` (`Get`, `GetLatest`, `GetLatestIncludingTerminating`, `ListTags`, `List`, `Upsert`, `DeleteByRef`) and makes `resource.Config.Store`, `resource.ApplyConfig.Stores`, the apply/delete cores, and the production admissions use it. `*v1alpha1store.Store` satisfies the interface unchanged, so every existing caller compiles as before.

Two small additions make a caller-owned store usable end to end:

- A store may deny an operation by returning `auth.ErrForbidden` or `auth.ErrUnauthenticated`. The handlers map those to 403 / 401 and the batch results label them `forbidden: ...`, matching what the `Authorize` hook produces. Existing stores never return these errors, so their behavior is unchanged.
- `types.AppOptions.V1Alpha1Stores` lets a downstream build supply a store for an extension kind. The app checks each value implements `resource.ObjectStore` at boot and merges it into the batch endpoint's store map, where it replaces any `V1Alpha1StoreTables` entry for the same kind. Route registration for the kind stays with the caller, as it does for every extension kind today.

Tests: a unit test drives an in-memory `ObjectStore` through the kind routes and `/v0/apply`, including the denial mapping on every route and in batch results. The existing integration tests keep their production stores and only change the map's element type.

# Change Type

/kind feature

# Changelog

```release-note
Extension kinds can supply their own resource store to the registry's generic CRUD and batch apply pipeline through the new resource.ObjectStore contract and the V1Alpha1Stores app option; stores may deny operations and the API returns 401 or 403 accordingly.
```

# Additional Notes

Validation on this commit: `go build ./...`, `go vet` with the unit and integration tags, the unit and Postgres integration suites for `pkg/registry/resource`, `internal/registry`, `internal/registry/api/router`, `internal/registry/api/handlers/v0/crud`, and `internal/client`, and `golangci-lint` with no findings.

No behavior change for existing kinds: every production store keeps its concrete type, and the new error mapping only fires on errors no existing store returns.


ilackarms:
Closing: the extension kind this was for can use the existing V1Alpha1StoreTables option with the per-kind authorizer and list-filter hooks, so the pipeline does not need a store interface seam. Leaving the branch in place in case the refactor is wanted on its own later.

